PR #2, opened 2016-11-21
more robust imports, better field handling - still need to include DOB in main file

jpfreeley:


JPF/.ipynb_checkpoints/mysql.skeleton-checkpoint.ipynb
```diff
@@ -0,0 +1,1710 @@
+{
```
chrishenry: I think `.ipynb_checkpoints` folders are for internal use only. The [default github python gitignore](https://github.com/github/gitignore/blob/master/Python.gitignore#L70) ignores them. Would you mind running a `git rm` on this folder?

I have [another branch](https://github.com/chrishenry/deltanyc-heatseek/commit/747efeb2acc21fbe059f5258f2465cb84daed2b3) where I'm adding the gitignore.

JPF/mysql.skeleton.ipynb
```diff
@@ -86,18 +86,18 @@
...
+     "end_time": "2016-11-21T10:20:48.596519",
+     "start_time": "2016-11-21T10:20:48.582484"
     },
     "collapsed": true
    },
    "outputs": [],
    "source": [
     "### AWS\n",
-    "#engine = create_engine('mysql+mysqlconnector://hsdbuser:hsdbpass@hsdb.cjjva3uq32na.us-west-2.rds.amazonaws.com:3306/heatseek', echo=False)\n",
+    "engine = create_engine('mysql+mysqlconnector://hsdbuser:156lafayette@hsdb.cjjva3uq32na.us-west-2.rds.amazonaws.com:3306/heatseek', echo=False)\n",
```
chrishenry: Let's pull this out for now. I'm working on [some changes](https://github.com/chrishenry/deltanyc-heatseek/commit/e5c5286d1f6527bed86a317e9b20038e2eed2bba#diff-c816bc3e049664cdf895812cd5ac0e0dR117) that will read db creds from environment instead of hardcoding.

JPF/mysql.skeleton.ipynb
```diff
@@ -119,24 +119,27 @@
...
+    "\n",
     "def guess_sqlcol(dfparam):    \n",
     "\n",
     "## GUESS AT SQL COLUMN TYPES FROM DataFrame dtypes. \n",
     "    \n",
     "    dtypedict = {}\n",
     "    for i,j in zip(dfparam.columns, dfparam.dtypes):\n",
     "        if \"object\" in str(j):\n",
-    "            dtypedict.update({i: sqlalchemy.types.NVARCHAR(length=1024)}) ##big field length for HPD violations description\n",
+    "            dtypedict.update({i: sqlalchemy.types.NVARCHAR(length=MAX_COLUMN_LENGTH)}) ##big field length for HPD violations description\n",
```
chrishenry, line 143: If I'm reading this right, we're going from 1024 -> 255 characters. Is that enough for the description?

JPF/mysql.skeleton.ipynb
```diff
@@ -167,19 +172,48 @@
...
     "        log.info(\"Why don't we save {} for next time\".format(output_pickle))\n",
     "        \n",
+    "    \n",
+    "    \n",
+    "    ## LET'S SEE IF THERE ARE COLUMNS TO TRUNCATE\n",
+    "    ## CLEAN COLUMN NAMES\n",
+    "    log.info(\"Time to clean up column names to lowercase no spaces etc...\")\n",
+    "    cols = [i.lower().replace(\" \",\"_\").replace(\"'\",\"\").replace(\"\\xe2\\x80\\x99\",\"\").\\\n",
+    "            replace(\"#\",\"num\").replace(\"&\",\"and\").replace(\"(\",\"\").\\\n",
+    "            replace(\")\",\"\") for i in df.columns]\n",
```
chrishenry, line 187: Cool. 

Alternatively, have you seen libs like [slugify](https://github.com/un33k/python-slugify)? Might be useful...

JPF/mysql.skeleton.ipynb
```diff
@@ -192,13 +226,241 @@
...
     "    log.info(\"Imported: {} rows\".format(df.shape[0]))\n",
+    "\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## DOB Violations\n",
+    "https://data.cityofnewyork.us/Housing-Development/DOB-Violations/3h2n-5cm9\n"
```
chrishenry, line 242: 👍 Super helpful to have the source URLs.

JPF/mysql.skeleton.ipynb
```diff
@@ -58,7 +58,7 @@
     "import pickle\n",
     "import logging\n",
     "\n",
-    "BASE_DIR = '/Users/jfreeley/Desktop/HeatSeek/HPD/Data Files/'\n",
+    "BASE_DIR = '/Users/jfreeley/Desktop/HeatSeek/'\n",
```
chrishenry: One super important thing to keep in mind is that when sharing this notebook with other people, no one else will be working in `/Users/jpfreely`. Options to make this more shareable are;

* Use `os.path.expanduser('~')` so the executing user's home directory is used.
* Make the home directory configurable via environment.

JPF/.ipynb_checkpoints/mysql.skeleton-checkpoint.ipynb
jpfreeley: i have git rm -r the .ipynb-checkpoints

JPF/mysql.skeleton.ipynb
jpfreeley: for the most part, not sure how much we will need the description .. would be better to have properly defined schema .. when I left it large, some tables with many varchar fields complained about too much data in the row. so far 255 has not caused issues .. but we need to be aware of potentially truncated data

JPF/mysql.skeleton.ipynb
jpfreeley: haven't seen slugify .. will investigate

JPF/mysql.skeleton.ipynb
jpfreeley: :-)

JPF/mysql.skeleton.ipynb
chrishenry: OK, that's good to know.

PR #3, opened 2016-11-22
implemented 'slugify' rather than .lower.replace...

jpfreeley:
Thanks for chris pointing out slugify as more elegant way to clean up field names.


chrishenry:
👍

PR #4, opened 2016-11-22
Added config.py skeleton and implementation for offline DB Creds

jpfreeley:


jpfreeley:
found errors in the scripts .. canceled PR for now.

PR #8, opened 2016-11-24
Big changes: load from URLS, pickle cleaned up, tested to HPD Violations

jpfreeley:
I made a slew of changes ... I've switched everything to Download URL. I'm paving the way to diff-only updates. 

I cleaned up pickle process by eliminating a pickle_file

11/26: I've added an SQL script which begins to create a "bad buildings" table from counts of various violations, permits, complaints, litigations. 

11/26: Added a few notes to NB and SQL

11/26: Added PubAdvWebScrape which scrapes Public Advocate's worst landlords buildings search page and captures, BuildingAddress, Landlord, #HPDViolations, #DOBViolations

11/26: Added PubAdv Worst Landlords to bad_buildings SQL file. 

11/27: Updated PubAdvWebScrape .. more descriptive VIO COUNT columns


JPF/PubAdvWebScrape.ipynb
```diff
@@ -0,0 +1,605 @@
...
+      "text/html": [
+       "<div>\n",
+       "<table border=\"1\" class=\"dataframe\">\n",
+       "  <thead>\n",
+       "    <tr style=\"text-align: right;\">\n",
+       "      <th></th>\n",
+       "      <th>ADDR</th>\n",
+       "      <th>LANDLORD</th>\n",
+       "      <th>HPDV</th>\n",
+       "      <th>DOBV</th>\n",
```
chrishenry: What do these columns represent?

JPF/PubAdvWebScrape.ipynb
```diff
@@ -0,0 +1,605 @@
...
+   ],
+   "source": [
+    "from selenium import webdriver\n",
+    "from bs4 import BeautifulSoup\n",
+    "import requests\n",
+    "import bs4\n",
+    "import pandas as pd\n",
+    "from bs4 import BeautifulSoup\n",
+    "from selenium.webdriver.common.desired_capabilities import DesiredCapabilities\n",
+    "from selenium import webdriver\n",
```
chrishenry: Selenium is super heavy for simple web scraping. Any reason we did just use something simple, like [requests](http://docs.python-requests.org/en/master/)?

JPF/bad_buildings.sql
```diff
@@ -0,0 +1,284 @@
...
+	hs_complaint_count
+select	c.buildingid,
+	count(*) as count
+from 
+	hpd_complaintsProb as cp
+left join 
+	hpd_complaints c
+on
+	c.complaintid = cp.complaintid
+left join
```
chrishenry, line 55: Any reason we went with the `LEFT JOIN` here? Typically when looking for a match in both tables, an `INNER JOIN` will give something closer to the result you're looking for.

JPF/bad_buildings.sql
```diff
@@ -0,0 +1,284 @@
...
+on 
+	c.buildingid = b.buildingid
+where 
+	majorcategoryid  in (12, 59, 65) or
+	minorcategoryid in (92, 196, 112, 375, 67, 323, 381, 106) or 
+	codeid in (800, 801, 802, 803, 804, 809, 1358, 1360, 2713, 2715, 2716, 
+	2833, 2490, 605, 616, 617, 629, 634, 640, 801, 1271, 2400, 2532, 2534, 
+	2661, 2724, 2764, 2767, 2769, 2794) 
+group by 
+	1;
```
chrishenry, line 66: Why the group here?

JPF/mysql.skeleton.ipynb
```diff
@@ -108,6 +107,11 @@
     "pw = os.environ['MYSQL_PASSWORD']\n",
     "db = os.environ['MYSQL_DATABASE']\n",
     "\n",
+    "user = 'root'\n",
+    "host = 'localhost'\n",
+    "pw = ''\n",
+    "db = 'heatseek'\n",
+    "\n",
```
chrishenry, line 114: Oh, were you not able to get environment variables working?

JPF/mysql.skeleton.ipynb
```diff
@@ -152,36 +156,38 @@
...
     "\n",
     "    else: \n",
-    "        log.info(\"Reading CSV from {} .. This may take a while...\".format(input_csv_file))\n",
+    "        log.info(\"Reading CSV from {} .. This may take a while...\".format(input_csv_url))\n",
     "        \n",
-    "        with open(input_csv_file, 'r') as input_csv: ## should just change to IF EXISTS rather than open()???\n",
-    "            df = pd.read_csv(input_csv_file , sep=sep_char, dtype=dtype_dict)\n",
-    "                \n",
-    "        log.info(\"Why don't we save {} for next time\".format(output_pickle))\n",
+    "        #SWITCHING TO URL DIRECTLY, CAN'T CHECK FOR LOCAL FILE ANYMORE\n",
```
chrishenry, line 177: Why can't we check for the local file?

chrishenry (commented):
Awesome! Couple questions.

JPF/mysql.skeleton.ipynb
```diff
@@ -152,36 +156,38 @@
...
     "    \n",
-    "    if os.path.isfile(input_pickle):\n",
-    "        log.info(\"Flagged load of PICKLE: {} = True\".format(input_pickle))\n",
+    "    if load_pickle: #IF FLAGGED TO LOAD PICKLE AS TRUE\n",
+    "        log.info(\"Flagged load of PICKLE: {} = True\".format(pickle_file))\n",
     "        \n",
-    "        with open(input_pickle, 'r') as picklefile:\n",
-    "            log.info(\"Begin OPEN {} Pickle: {}\".format(input_pickle, datetime.datetime.now()))\n",
-    "            log.info(\"Great we have a pickle file...Loading from {}\".format(input_pickle))\n",
+    "        with open(pickle_file, 'r') as picklefile:\n",
```
chrishenry, line 169: Without a check that the file exists, this will error on the initial run of the notebook.

Perhaps change the logic above to 

```python
if load_pickle and os.path.isfile(pickle_file):
````

That will load the pickle if it exists, but fall back to loading via CSV if not.

File: JPF/PubAdvWebScrape.ipynb (hunk 1 above)
Comment (jpfreeley) [reply JPF/PubAdvWebScrape.ipynb]: those are the data that the Pub Advocate site provides. 
the Address of the Bad Landlord Building
the name of the Bad Landlord
the #of HPD Violations the Public Adv has assigned, based on their method, to the BUILDING
the #of DOB Violations the Public Adv has assigned, based on their method, to the BUILDING
see: http://landlordwatchlist.com/search.html
see: http://advocate.nyc.gov/Methodology

File: JPF/PubAdvWebScrape.ipynb (hunk 2 above)
Comment (jpfreeley) [reply JPF/PubAdvWebScrape.ipynb]: I can try with "requests", however considering I can't see the data in a chrome "view page source" I think selenium *may* be warranted.

File: JPF/bad_buildings.sql (hunk 3 above)
Comment (jpfreeley) [reply JPF/bad_buildings.sql:55]: i wanted to always maintain the FULL  LIST of buildings with the NULLS for those without counts so I was never dealing with a subset. perhaps overkill and performance hit.
probably mostly due to my lack of experience. in the final table, i didn't want to end up with a list significantly smaller than the original building list.

File: JPF/bad_buildings.sql (hunk 4 above)
Comment (jpfreeley) [reply JPF/bad_buildings.sql:66]: if i don't group on buildings, it won't allow me to count(*) the records.

File: JPF/mysql.skeleton.ipynb (hunk 5 above)
Comment (jpfreeley) [reply JPF/mysql.skeleton.ipynb:114]: i did but I need to restart my ipython process to re-read the env when i switch between AWS/Local .. easier this way where AWS is in hidden ENV vars and I can easily overwrite with locals .. i know this is probably bad form, I didn't intend for them to get committed.

File: JPF/mysql.skeleton.ipynb (hunk 6 above)
Comment (jpfreeley) [reply JPF/mysql.skeleton.ipynb:177]: local file no longer presumed to be used. Can check, but everything is now live URL , no local CSVs.

File: JPF/mysql.skeleton.ipynb (hunk 7 above)
Comment (jpfreeley) [reply JPF/mysql.skeleton.ipynb:169]: yes, I've been trying to get my head around best way to manage this. I do want control (at least during testing) of whether to load pickle or not .. after development, the pickles will probably go away. i'd like to move to a "get newest only after initial base load" from Socrata URL via API URL .. this is a problem with DOB data.

File: JPF/PubAdvWebScrape.ipynb (hunk 1 above)
Comment (chrishenry) [reply JPF/PubAdvWebScrape.ipynb]: Ah, got it. It'd be super helpful if you just clarified the last 2 columns. Maybe something like `HPV Violation Count`?

File: JPF/PubAdvWebScrape.ipynb (hunk 2 above)
Comment (chrishenry) [reply JPF/PubAdvWebScrape.ipynb]: Oh...I see. Taking a look, it looks like just have all the data in a [javascript file](http://landlordwatchlist.com/search.js). So, you could do something like this;

```python
import requests

r = requests.get("http://landlordwatchlist.com/search.js")
data = json.loads(r.text.split(";")[0].replace("var markers = ", ''))
```

JPF/bad_buildings.sql
chrishenry: Ah, got it. I'm not actually familiar with the `GROUP BY {index}` syntax. (Had to look it up). 

For others, it may be helpful if you clarify the syntax to `group by c.buildingid`

JPF/mysql.skeleton.ipynb
chrishenry: Got it. No worries. It's not bad form to make changes like that while working locally, but just don't forget to pull them out when submitting code. Otherwise the next person who pulls will be quite confused as to why their config isn't working.

JPF/mysql.skeleton.ipynb
chrishenry: I'd argue that we should try to maintain a workflow of download -> pickle -> load, with checks that will skip the download and pickle steps if on disk.

The biggest reason being that opening the file via network every time is pretty inefficient. Any issue in pickling or loading and you have re-download and re-pickle again. 

Also, while I'm not super familiar with pandas internals, opening via http probably doesn't give us the option to read the csv in chunks, resulting in a more memory efficient load process. For example, I've been watching this process for ~4 hours now.

JPF/bad_buildings.sql
chrishenry: Ah, got it. Do you intend for this to be a step in a series of data transformations?

JPF/mysql.skeleton.ipynb
chrishenry: Right. Is there a way to pass flags or parameters into ipython notebook cells?

Ideally, you'd set up the logic to default to using what's on disk if exists, but override the behaviour if a flag is set.

JPF/bad_buildings.sql
jpfreeley: funny, i thought this syntax was much more flexible and was glad to have discovered it.

JPF/bad_buildings.sql
jpfreeley: not sure exactly your question . we can discuss further later ..I may be asking others to start jumping in on some of this, and whether my direction makes sense.

JPF/mysql.skeleton.ipynb
jpfreeley: 4 hours? the entire load from scratch should only take a bit (45-60min) over reasonable wifi. 

i'm happy to heed your direction. for the most part, i imagine this should be a "base load once" process, in which case they will not likely have either the CSV nor the pickle, with future loads only being from "newest" diffs if possible. happy to further discuss.

JPF/mysql.skeleton.ipynb
jpfreeley: not afaik, can't pass ipython params. 

what's on disk will nearly always be stale (DOB and 311) are daily updates. HPD monthly, PubAvd Quarterly

JPF/mysql.skeleton.ipynb
jpfreeley: yep.

PR #9, opened 2016-11-28
add directory containing rails app

LKWLaLa:
@chrishenry 

I've moved my rails files into the directory 'rails-app.'  All seems to be working locally!


chrishenry:
👍

PR #12, opened 2016-12-04
Most recent changes

LKWLaLa:
@chrishenry 

A few notes:

1.  I didn't repair the database.yml file - simply changed the order of listings so that I could access my database.  Might want to take a peek at that.  :)

2.  If you run rails s, there is a front end now!  I'm still working on the main menu feature, and I'll submit a pull when I get the dropdown menu and navigation to a particular properties page working.  (I'm guessing maybe two hours or so from now.)

3.  I still need to write my README, so I'll get working on that too.

Will be reachable if you want to discuss anything.  Thanks!!


LKWLaLa:
@chrishenry 

Auto-suggest working on home page for address input.

LKWLaLa:
Actually, my auto-suggest is not currently supported in safari.  Chrome, firefox works.  May implement this another way.

chrishenry:
I'm going to merge this, as I need to get into your most recent schema files to start massaging data into your schema.

LKWLaLa:
@chrishenry Ok, sounds good!  If you want to check out the front end, use Chrome for now.

PR #15, opened 2016-12-05
Implement typeahead

LKWLaLa:
@chrishenry 

New implementation for the auto-suggest on home page.  Now using Angular UI Bootstrap directive Typeahead, rather than HTML 5 datalist.  This fixes problem of compatibility with Safari.


chrishenry:
Hey, after pulling in the first set of `properties`, I think we're going to have to move the actual search to the server side. There's probably going to be ~800k addresses we're searching, and doing it on the client side will *not* be performant.

LKWLaLa:
Crap....ok.  Hm.  Would you mind just pulling in my latest updates and trying the front end with the new implementation once?  It's probably terribly slow, but I'd just like to check it before I start over again.

LKWLaLa:
Never mind.... you're right.  I just seeded my file with 4K addresses, and it significantly slows down.

LKWLaLa:
Ok, I'm going to try to implement the Google maps API autocomplete class, and then we can send an ajax call to the backend, and query for one full address.   Does that sound like a good idea?

chrishenry:
Yep, that sounds like a good bet. It'll give the user a good search experience, and us structured address data we can use to match against our properties table.

LKWLaLa:
Ok, I'll work on that.

LKWLaLa:
@chrishenry 

Implemented Google Places API.

Question:  Can we format data in such a way that I can query with an address in the format below? 

455 W 34th St, New York, NY 10001, USA

Noting - Google uses a shortened "W 34th St," rather than "West 34th Street" or "W. 34th St."  etc.  Thoughts?

LKWLaLa:
Actually, here are the details of the JSON object returned for a particular Google maps "place" object:

https://developers.google.com/places/web-service/details

If it would be more standardized to query using a particular attribute other than the formatted_address, let me know what you think makes the most sense.

chrishenry:
Nice!

I'd argue you should send up the entire `address_components` object for now. It'll give some flexibility in creating the query.

LKWLaLa:
I'm not sure how to query using the entire address components object in Rails.  Would you know how to do that?

Also, I did something funny in the initial set up of the complaint_311s table, and the naming conventions are confused.  If I drop this table and re-do, will that create problems on your end? (I'm currently creating serializers for the rest of the tables, so I can add onto the endpoints.)

chrishenry:
Hey, to clarify sending up `address_components`; we can pick and choose what `components` we'll use to search our data. It'll probably end up looking something like this;

```ruby
street_address = get_long_street_number(components) + " " + get_long_route(components)
zipcode = get_long_postal_code(components)

@properties = Property.where("street_address LIKE '%:street_address%' AND zip = :zipcode",
    {:street_address => street_address, :zipcode => zipcode})
```

Obvi, we'll need to implement some helpers to handle the address_components objects (or find an existing lib), but that's the general idea.

Regarding the 311 table, yes, you can absolutely drop and re-create. (Rails' `db:migration` task should handle this for us.)

angular-rails-app/app/views/layouts/application.html.erb
```diff
@@ -5,6 +5,10 @@
   <%= stylesheet_link_tag    'application', media: 'all'%>
   <%= javascript_include_tag 'application'%>
   <%= csrf_meta_tags %>
+
+
+    <script src="https://maps.googleapis.com/maps/api/js?v=3.exp&sensor=false&libraries=places"></script>
```
chrishenry: Will we need an API key for this?

LKWLaLa: It seems that version 3 does not currently need a key, but it might be a good idea to use one anyhow:   http://stackoverflow.com/questions/35890582/google-maps-api-warning-noapikeys

LKWLaLa:
Regarding dropping and re-creating the 311 table - ok great. I just wanted to make sure recreating it wouldn't cause any issues on your end (with the _r prefix or anything else you have built.)  Also, Andrew pointed out to me this evening, that the correct acronym is hpd - not hdp - as I have named all of my tables.  So, I will be changing that as well.

LKWLaLa:
On the note of the `address_components` object... the challenge I ran into when I explored drilling into it, is that, for example: `address_components[6]` does not represent the same thing for each address.  (Which is a bit annoying.)  So if we want to find a state or a zipcode, we can't depend on it being located in the same place each time.  Some addresses have more information than others, and the objects have more properties.  

That is what led me to considering the `formatted_address` property of the object.  Do you think it would be effective to return that, and then parse the string into the various fields we are querying?  Would using `LIKE` in combination with `AND` be specific enough to return the row/instance we are looking for?

chrishenry:
Yep, the numeric index is insufficient to find the component you're looking for. (Maybe at some point @rdlester can shed some light on why that structure is the way it is.)

I typically implement a helper where you can pass in the `address_components` array and the component you want. Here's an example of the last one I did, in JS.

```javascript
  googResultGetComponent: function(result, component, len) {
    var x = $.grep(result, function(c){ return c.types.indexOf(component) >= 0; });
    return x[0][len || 'long_name'];
  }
```

w/r/t whether the query will be specific enough, I'm not honestly sure, we'll have to play with it.

rdlester:
@jpfreeley knows more about the 311 fields than I do.

LKWLaLa:
@chrishenry   Ah, thank you!  Ok, I see what you mean.  And actually, your helper in JS looks great...  seems like I could possibly use that verbatim.

LKWLaLa:
Serializers and naming conventions fixed and working.  If you navigate to `localhost:3000/properties/:id` you should see the full json object for a property, along with the hpd_violations, 311 info, owners, etc.

You can also navigate to `localhost:3000/owners/:id` and see the owner information along with all associated properties.

In oder to test the database address query, I think I'm at the point where I'll need to get docker running on my end, as I can't test the query on fake data.  

I used a slightly different approach for the helper to parse the google place json  (thanks so much for your explanations though - that was helpful!) - that part is now working in Angular.  Moving on to the query on the Rails side.

rdlester:
Ha, I really misinterpreted the comment about address components.

The true docs appear to be at: https://developers.google.com/maps/documentation/geocoding/intro#Types
which was several links removed from the autocomplete API docs, so it's not easy to find!

It looks like address_components is simply the address broken up into each individual word. It also looks like address_components contains (somewhere) a types[] array indicating the type of each individual address component. The docs are not very clear about any of this, however, sorry.

rdlester:
Oh, here we go, the real explanation of address components:
https://developers.google.com/maps/documentation/geocoding/intro#Results

LKWLaLa:
@rdlester Thanks!  :)

@chrishenry  The Rails query is working, and returning a Property object when I use an exact match on zipcode.  However, I am struggling with how to query for the street address.  Using `Property.where("street_address LIKE '%:street_address%'` only allows for wild cards on each end.  However, this won't return a result when the difference is something like `50 W. 52nd St` and `50 West 52nd st`.  Anyone have thoughts?

chrishenry:
We may want to take a look at [fuzzy matching](https://github.com/seamusabshere/fuzzy_match) or FULLTEXT indexing for allowing for some leeway when doing these searches.

I'm going to merge this in for now.

rdlester:
Re: addresses, we will probably need to normalize them.
http://www.rubygeocoder.com/ ?

PR #17, opened 2016-12-06
Hid away the postgres files.

jpfreeley:


chrishenry:
Good call, these will probably be useful at some point.

PR #18, opened 2016-12-06
Joined rent stabilization script.

rdlester:
Added arg for switching off default primary key creation.

### DIFF
--- a/JPF/rent-stab.py
+++ b/JPF/rent-stab.py
@@ -27,7 +27,7 @@ TAXBILLS_KEY = 'taxbills_joined'
 
 taxbills_dtype_dict = {
     "borough": "object",
-    "ucbbl": "int64",
+    "ucbbl": "int32",
     "2007uc": "float64", "2007est": "object", "2007dhcr": "object", "2007abat": "object",
     "2008uc": "float64", "2008est": "object", "2008dhcr": "object", "2008abat": "object",
     "2009uc": "float64", "2009est": "object", "2009dhcr": "object", "2009abat": "object",
@@ -94,6 +94,8 @@ taxbills_keep_cols = [
     "2015abat"
 ]
 
+taxbills_primary_key = "ucbbl"
+
 
 def main(argv):
 
@@ -117,7 +119,7 @@ def main(argv):
     taxbills_description = "Taxbills Rent Stabilization"
     taxbills_pickle = taxbills_joined_dir + '/taxbills.pkl'
     taxbills_sep_char = ","
-    taxbills_table_name = "rent-stabilization"
+    taxbills_table_name = "rent_stabilization"
     taxbills_load_pickle = args.LOAD_PICKLE
     taxbills_save_pickle = args.SAVE_PICKLE
     taxbills_db_action = 'replace' ## if not = 'replace' then 'append'
@@ -139,7 +141,7 @@ def main(argv):
                 taxbills_date_time_columns,
                 taxbills_chunk_size,
                 taxbills_keep_cols,
-                set_default_primary_key=False
+                primary_key=taxbills_primary_key
                 )
 
 
--- a/JPF/rent-stab.py
+++ b/JPF/rent-stab.py
@@ -1,0 +1,147 @@
+#!/usr/bin/env python
+
+import argparse
+import os
+import os.path
+import sys
+import logging
+
+
+from sqlalchemy import create_engine
+
+from utils import *
+
+mkdir_p(BASE_DIR)
+
+logging.basicConfig(format='[%(asctime)s] {%(pathname)s:%(lineno)d} %(levelname)s - %(message)s',
+    datefmt='%H:%M:%S',
+    stream=sys.stdout,
+    level=logging.INFO)
+log = logging.getLogger(__name__)
+
+"""
+Taxbills Rent Stabilization joined count import
+"""
+
+TAXBILLS_KEY = 'taxbills_joined'
+
+taxbills_dtype_dict = {
+    "borough": "object",
+    "ucbbl": "int64",
+    "2007uc": "float64", "2007est": "object", "2007dhcr": "object", "2007abat": "object",
+    "2008uc": "float64", "2008est": "object", "2008dhcr": "object", "2008abat": "object",
+    "2009uc": "float64", "2009est": "object", "2009dhcr": "object", "2009abat": "object",
+    "2010uc": "float64", "2010est": "object", "2010dhcr": "object", "2010abat": "object",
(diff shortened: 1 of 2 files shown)


JPF/utils.py
```diff
@@ -104,6 +104,7 @@ def hpd_csv2sql(description, input_csv_url, sep_char,\
         max_col_len - length to truncate cells in truncate_columns to
         date_format - expected format of dates in the table
         csv_chunk_size - if defined, the number of rows to process from the CSV at a time.
+        set_default_primary_key - if True, adds an ID column to serve as primary key.
```
rdlester: Presumably, there should be an arg allowing callers to specify a column name to use as the primary key (ucbbl in this case), but I'm not sure what the SQL syntax is for the ALTER TABLES call below.

JPF/utils.py
```diff
@@ -239,7 +242,7 @@ def send_df_to_sql(df, log, description, table_name, db_action, sql_chunk_size,
     df.to_sql(name=table_name, con=conn, if_exists=action,\
               index=False, chunksize=sql_chunk_size, dtype=outputdict)
 
-    if db_action == 'replace':
+    if db_action == 'replace' and set_default_primary_key:
         conn.execute("ALTER TABLE %s ADD id INT PRIMARY KEY AUTO_INCREMENT;" % table_name)
```
rdlester: Presumably, I can just replace "id" here with any column name (that uses INT type), right?

JPF/rent-stab.py
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env python
```
chrishenry, line 1: Rent always makes me stabby, too. 😀

JPF/rent-stab.py
```diff
@@ -0,0 +1,147 @@
...
+    "ucbbl": "int64",
+    "2007uc": "float64", "2007est": "object", "2007dhcr": "object", "2007abat": "object",
+    "2008uc": "float64", "2008est": "object", "2008dhcr": "object", "2008abat": "object",
+    "2009uc": "float64", "2009est": "object", "2009dhcr": "object", "2009abat": "object",
+    "2010uc": "float64", "2010est": "object", "2010dhcr": "object", "2010abat": "object",
+    "2011uc": "float64", "2011est": "object", "2011dhcr": "object", "2011abat": "object",
+    "2012uc": "float64", "2012est": "object", "2012dhcr": "object", "2012abat": "object",
+    "2013uc": "float64", "2013est": "object", "2013dhcr": "object", "2013abat": "object",
+    "2014uc": "float64", "2014est": "object", "2014dhcr": "object", "2014abat": "object",
+    "2015uc": "float64", "2015est": "object", "2015dhcr": "object", "2015abat": "object",
```
chrishenry, line 39: Never seen this style, but it kinda makes sense.

JPF/utils.py
```diff
@@ -84,11 +84,11 @@ def guess_sqlcol(dfparam, max_col_len):
...
-def hpd_csv2sql(description, input_csv_url, sep_char,\
-            table_name, dtype_dict, load_pickle, save_pickle, \
-                pickle_file, db_action, truncate_columns, date_time_columns,\
-               sql_chunk_size, keep_cols, max_col_len=255, date_format=None, \
-               csv_chunk_size=None):
+def hpd_csv2sql(description, input_csv_url, sep_char,
+            table_name, dtype_dict, load_pickle, save_pickle,
+                pickle_file, db_action, truncate_columns, date_time_columns,
+               sql_chunk_size, keep_cols, max_col_len=255, date_format=None,
+               csv_chunk_size=None, set_default_primary_key=True):
```
chrishenry: I wonder if `False` would be a better default, to stay in line with existing behaviour. I don't think we're actually using the PK anywhere. It could def be useful, but we'd need to collectively figure out a use for it.

JPF/utils.py
chrishenry: The intent was to add a straightforward `INT` primary key, independent of the content of the table.

JPF/utils.py
chrishenry: Yesh. It should be that easy.

After a quick google, it seems there is a newer, clearer syntax for [adding indexes in MySQL](http://dev.mysql.com/doc/refman/5.7/en/create-index.html). We may want to cut over to that, since it's cleared.

jpfreeley:
Not sure how to add to your comment above .. but the "new mysql" create index syntax is actually to match the postgres syntax https://www.postgresql.org/docs/9.5/static/sql-createindex.html .. we are now 1 step closer ..

JPF/utils.py
rdlester: Dang, there's a new twist - pandas.to_sql seems to convert all ints (even int32) to BIGINT sql type, which can't be used as a primary key.

There is an option on to_sql to pass in a dict mapping columns to SQL types, but that seems like a pain to fill out. It would solve this problem however.

rdlester: I was wrong about that, there was another weird error in SQL statements (and MySQL does not even attempt to hint at what is wrong?).

Regardless, I set things up to automatically convert the given column to INT type when setting as PRIMARY KEY.

rdlester:
Setting up Indexes should totally be a part of the scripts, but I'll leave that for other PRs.

jpfreeley:
are you sure it's not the guess_sqlcol() function in utils that is setting
the bigint?

rdlester:
Oh, I didn't notice the guess_sqlcol() func, that is probably it.

Anyways, we probably do want BIGINTs in the majority of cases. Might be
nice to set int32 -> INT just to have the capability, but we'd have to go
back through all the data -> dtype dicts again to evaluate which can be
which, which maybe isn't super important.

As mentioned, I would like to set up indexes right in the scripts. As a
MySQL newbie, I don't totally understand the nuances of indexes, though; we
should talk about this tonight.

JPF/rent-stab.py
rdlester: Well, the mixed formatting is definitely not kosher, but I think it works for the better here.

rdlester: unless we want to agree on a Python linter

rdlester:
Merged in the SKIP_IMPORT flag. Took the default Primary Key out, moved it into the rent-stab.py. Other scripts will have to set their own Primary Keys (presumably to BBL).

rdlester:
I think this is set to go, now that the Primary Key stuff has been resolved, any other issues left that need taking care of?

JPF/rent-stab.py
chrishenry: In a few iterations, maybe...let's get an end-to-end integration working well, first.

chrishenry:
👍

PR #19, opened 2016-12-06
Drop cols via usecols arg on read_csv

rdlester:
Will prevent unused columns from ever being loaded into memory beyond during parsing, hopefully improving memory usage and time.

(I didn't verify every script still worked, please double check.)


rdlester:
I checked that the scripts all work.

JPF/utils.py
```diff
@@ -134,10 +134,11 @@ def hpd_csv2sql(description, input_csv_url, sep_char,\
             return
 
         # stream csv into chunked dataframes
-        log.info("Streaming CSV from {} ...".format(input_csv_url))
+        log.info("Streaming {} CSV fields from {} ...".format(
+            len(keep_cols), input_csv_url))
 
         csv_chunks = pd.read_csv(input_csv_url, sep=sep_char, dtype=dtype_dict,\
-                encoding='utf8', chunksize=csv_chunk_size)
+                encoding='utf8', chunksize=csv_chunk_size, usecols=keep_cols)
```
chrishenry, line 141: Sweet.

For those following along, here's the docs for [pandas.read_csv](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.read_csv.html)

jpfreeley: awesome .. i should have read those docs weeks ago .. at least I made it easy to utilize the usecols ... does it matter that my keep_cols is after lowered slugified cols?

rdlester: Yeah, unfortunately - note the other file changes in this PR.

PR #21, opened 2016-12-07
Start of rake tasks to pull data into rails tables

chrishenry:


chrishenry:
Going to merge this in to get Docker, and rails app on the same page.

PR #23, opened 2016-12-08
fix skip_import

rdlester:
whoops


rdlester:
Just gonna merge this now, it breaks everything.

PR #24, opened 2016-12-10
import-dob-permits/violations

rdlester:


JPF/import-dob-permits.py
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+
+perm_dob_df_keep_cols = [
+        'BOROUGH',
+        'Bin #',
+        'House #',
+        'Street Name',
+        'Job #',
+        'Job doc. #',
```
rdlester, line 89: Frak these column names, seriously, wtf with the period and not capitalizing every word

jpfreeley:
I abhor #DO.b data.

JPF/import-dob-permits.py
chrishenry: Points for `frak`

JPF/import-dob-permits.py
```diff
@@ -0,0 +1,179 @@
...
+    else:
+        log.info("DOB Permits exists, moving on...")
+
+    perm_dob_description = 'DOB Permits'
+    perm_dob_pickle = dob_permits_dir + '/df_dob_permit.pkl'
+    perm_dob_sep_char = ","
+    perm_dob_table_name = "dob_permits"
+    perm_dob_load_pickle = args.LOAD_PICKLE
+    perm_dob_save_pickle = args.SAVE_PICKLE
+    perm_dob_db_action = "replace"
```
chrishenry: We should use the `--update-or-replace-db` [flag](https://github.com/chrishenry/deltanyc-heatseek/blob/master/JPF/utils.py#L38)

chrishenry (commented):
Other than the flag, lgtm.

JPF/import-dob-permits.py
rdlester: Done.

chrishenry:
👍

PR #25, opened 2016-12-10
rename 311 to match other scripts

rdlester:


rdlester:
Just gonna merge this, it's such a simple change.

PR #29, opened 2016-12-12
Aep import

rdlester:
Adds name to first AEP column in csv file; moves CSV to data dir.

### DIFF
--- a/data-imports/import-aep.py
+++ b/data-imports/import-aep.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python
+
+import argparse
+import os
+import os.path
+import sys
+import logging
+
+from utils import *
+
+mkdir_p(BASE_DIR)
+
+logging.basicConfig(format='[%(asctime)s] {%(pathname)s:%(lineno)d} %(levelname)s - %(message)s',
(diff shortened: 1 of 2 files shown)


rdlester:
Also will blow up without the fix-args PR.

data-imports/import-aep.py
```diff
@@ -0,0 +1,51 @@
...
+logging.basicConfig(format='[%(asctime)s] {%(pathname)s:%(lineno)d} %(levelname)s - %(message)s',
+    datefmt='%H:%M:%S',
+    stream=sys.stdout,
+    level=logging.INFO)
+log = logging.getLogger(__name__)
+
+table_name = 'aep_list'
+
+def main(argv):
+    parser = argparse.ArgumentParser(description='Import hpd buildings dataset.')
```
chrishenry: Looks like a copy-pasta.

rdlester: yeah, i can clean up the abstractions to lessen this sort of thing in a different PR

rdlester: done

rdlester:
Doesn't seem like there are any other comments; merging

PR #31, opened 2016-12-12
Beginnings of Data Sci Model Development

jpfreeley:


JPF/bad_buildings.sql
```diff
@@ -348,9 +391,35 @@ drop table if exists bad_buildings;
...
+	if (pnyc.unitsres != 0, (all_311_hpd_count / pnyc.unitsres) ,0) all_311_hpd_count_per_resunits,
+	if (pnyc.unitsres != 0, (dob_violation_cnt / pnyc.unitsres) ,0) dob_violation_cnt_per_resunits,
+	if (pnyc.unitsres != 0, (hs_permit_cnt / pnyc.unitsres) ,0) hs_permit_cnt_per_resunits,
+	if (pnyc.unitsres != 0, (tbcs.unitsstab2007 / pnyc.unitsres) ,0) rentstab2007_per_resunits,
+	if (pnyc.unitsres != 0, (tbcs.unitsstab2015 / pnyc.unitsres) ,0) rentstab2015_per_resunits,
+	if (pnyc.unitsres != 0, (tbcs.diff / pnyc.unitsres) ,0)  diff_per_resunits, 
+	if (pnyc.unitsres != 0, (cvA.classA_cnt / pnyc.unitsres) ,0)  classA_cnt_per_resunits, 
+	if (pnyc.unitsres != 0, (cvB.classB_cnt / pnyc.unitsres) ,0)  classB_cnt_per_resunits, 
+	if (pnyc.unitsres != 0, (cvC.classC_cnt / pnyc.unitsres) ,0)  classC_cnt_per_resunits, 
+	if (pnyc.unitsres != 0, (cvI.classI_cnt / pnyc.unitsres) ,0)  classI_cnt_per_resunits,
```
rdlester, line 450: Just gonna note this here, as it seems like the best place - we should be careful about including the violations as features in the model if we use the worst landlord list and AEP as the objective, as the direct dependence between the two quantities might lead to a model that simply reproduces the worst landlord list.

jpfreeley:
So this is the general plan at this time ..( however, it's falling flat due
to inability to get matching data (unit counts, b&c counts, dob_vio counts)
to the PA-WLL ....)
1) Calculate a WLL score for all buildings (which aren't HPD Demolished :-)
) -- bb_score
2) Binary 0 good/1 bad as per the thresholds of WLL methodology  --
bb_target
3) Discard the features that went into making the WLL score (b_count,
c_count, dob_v_count, unitsres) and throw out the bb_score
4) Use the remaining features to predict into the bb_target binary.

rdlester:
Hmm. The only goofiness with that model is that all the data is (technically) labelled, so we've just got training / validation data, not test data. You could, say, only calculate bb_target for half the data and predict values for the other half, but then we'd only have unbiased scores for half the data.

I do certainly think we can learn something from this, but we have to keep that limitation in mind and probably shouldn't report these scores publicly ... ?

rdlester:
Just for the record, I'm not totally sure how to get around the training / test split problem.

jpfreeley:
It was discussed last night that the scores should not be posted anywhere
at this time.

i've run my method through Logistic Regression and Gradient Boosting nd
have accuracy scores around 0.80 with a train/test -split of 33%.

I'm very good at predicting good buildings but not great at predicting bad
buildings. See attached confusion matrix and ROC curve.

rdlester:
I apologize for not reviewing this more earlier. I don't really know SQL and that SQL file is especially messy / a little confusing. I'll try to look through and give comments / ask for clarification on things now, though.

JPF/bad_buildings.sql
```diff
@@ -8,17 +8,18 @@
...
 # !!! Needs better documentation
-# !!! Ends up with 2000 MORE BUILDINGS THAN EXIST IN hpd_buildings .. likely dirty addresses .. FIRST()??
-### This finds them (1955 rows)
+# !!! Ends up with 26 MORE BUILDINGS THAN EXIST IN hpd_buildings .. likely dirty addresses .. FIRST()??
+### This finds them (26 rows)
 ### select hbaddress, buildingid, count(buildingid) count from bad_buildings group by 1,2 having count > 1;
 
 
 ## BELOW REQUIRES THAT NO INDEXES/KEYS PRE-EXIST ON THE TABLES
 ## AT THE BOTTOM OF THIS FILE IS CODE FOR REMOVING THE FOLLOWING INDEXES/KEYS
```
rdlester, line 19: File-wide comment: Put explanatory comments before each major operation or set of operations. Someone who knows SQL could probably figure out what's going on / what the point of each line is, but it'd take a while + a lot of work. It's especially difficult to keep track of things given how large this file is.

JPF/bad_buildings.sql
```diff
@@ -32,15 +33,38 @@ alter table hpd_registrations add index(registrationid);
 alter table hpd_registrationContact add index(registrationcontactid);
 alter table hpd_registrationContact add index(type);
 alter table hpd_registrationContact add index(registrationid);
-alter table `pubadv_worst_landlords` add index(bin);
+alter table pubadv_worst_landlords add index(bin);
 alter table pluto_nyc add index(bbl);
 alter table tb_changes_summary add index(ucbbl);
 alter table hpd_aep_list add index(full_addr);
 
+delete from hpd_buildings where buildingid = 814731; #FICTICIOUS?? OVERSIZED BBL .. DELETE
```
rdlester, line 41: This isn't a BBL...?

JPF/bad_buildings.sql
```diff
@@ -32,15 +33,38 @@ alter table hpd_registrations add index(registrationid);
...
 
 delete from dob_violations where boro not in ('1','2','3','4','5');
```
rdlester, line 43: The entries with invalid boro ids still appear to be valid addresses. One or two of them are ambiguous (an address that occurs in multiple boroughs), but the others are uniquely determined by address. Is it worth doing special cleaning for those entries? (This is a Q for Chris and Lindsey as well - do we want to be responsible for actually correcting mistakes in city data and maintaining those corrections, or should we just drop those entries to reduce the workload?)

JPF/bad_buildings.sql
```diff
@@ -32,15 +33,38 @@ alter table hpd_registrations add index(registrationid);
...
+	
+delete from hpd_buildings 
+where 
+	bin < 1000000 or 
+	bin > 5999999 or 
+	bin = 1000000 or 
+	bin = 2000000 or 
+	bin = 3000000 or 
+	bin = 4000000 or 
+	bin = 5000000;
```
rdlester, line 67: It sucks that SQL doesn't have an easy way to define and reuse functions. Stored procedures would be nice but it looks like you have to use dynamic SQL to parameterize the table name.

This is a major advantage to constructing the statements in Python - you can easily reuse and format statements. This should go in clean_utils.py as a method, for sure - would you mind adding it + adding it to the cleaning for these tables?

(Also - aren't there other tables with BBLs?)

JPF/bad_buildings.sql
```diff
@@ -32,15 +33,38 @@ alter table hpd_registrations add index(registrationid);
...
 alter table tb_changes_summary add index(ucbbl);
 alter table hpd_aep_list add index(full_addr);
 
+delete from hpd_buildings where buildingid = 814731; #FICTICIOUS?? OVERSIZED BBL .. DELETE
 
 delete from dob_violations where boro not in ('1','2','3','4','5');
 ALTER TABLE dob_violations CHANGE `boro` `boro` INT(5)  NULL  DEFAULT NULL;
 alter table dob_violations add index(boro);
+alter table dob_violations add index(bin);
+alter table dob_permits add index (borough);
```
rdlester, line 47: Bump these indexing statements up to the top block of index statements, just so all the indexes are in a single place?

JPF/bad_buildings.sql
```diff
@@ -32,15 +33,38 @@ alter table hpd_registrations add index(registrationid);
...
-alter table `pubadv_worst_landlords` add index(bin);
+alter table pubadv_worst_landlords add index(bin);
 alter table pluto_nyc add index(bbl);
 alter table tb_changes_summary add index(ucbbl);
 alter table hpd_aep_list add index(full_addr);
 
+delete from hpd_buildings where buildingid = 814731; #FICTICIOUS?? OVERSIZED BBL .. DELETE
 
 delete from dob_violations where boro not in ('1','2','3','4','5');
 ALTER TABLE dob_violations CHANGE `boro` `boro` INT(5)  NULL  DEFAULT NULL;
```
rdlester, line 44: What is this line doing? Isn't boro already an INT? (Or did it get encoded as varchar)

JPF/bad_buildings.sql
```diff
@@ -32,15 +33,38 @@ alter table hpd_registrations add index(registrationid);
...
+where 
+	bin < 1000000 or 
+	bin > 5999999 or 
+	bin = 1000000 or 
+	bin = 2000000 or 
+	bin = 3000000 or 
+	bin = 4000000 or 
+	bin = 5000000;
 
 drop table if exists hpd_registrationContact_d;
```
rdlester, line 69: Comment - what is hpd_registrationContact_d?

Also - a simpler way to delete duplicate rows is to create a temporary unique index, via this Stack Overflow Q (answered by Chris!): https://stackoverflow.com/questions/3311903/remove-duplicate-rows-in-mysql

JPF/bad_buildings.sql
```diff
@@ -56,6 +80,16 @@ drop table if exists boro_lookup;
 ################
 create temporary table boro_lookup select boro, boroid from hpd_buildings group by 1,2;
```
rdlester, line 81: Is "GROUP BY 1,2" valid? Shouldn't it be "boro, boroid"? What do these numbers refer to?

JPF/bad_buildings.sql
```diff
@@ -56,6 +80,16 @@ drop table if exists boro_lookup;
 ################
```
rdlester, line 80: Comment - What is the point of this table?

JPF/bad_buildings.sql
```diff
@@ -56,6 +80,16 @@ drop table if exists boro_lookup;
...
 create temporary table boro_lookup select boro, boroid from hpd_buildings group by 1,2;
 alter table boro_lookup add index(boroid);
+alter table boro_lookup add index(boro);
+
+################
```
rdlester, line 85: Comment.

JPF/bad_buildings.sql
```diff
@@ -71,7 +105,7 @@ left join
 on
 	c.complaintid = cp.complaintid
```
rdlester, line 106: What is "c"? If it's hpd_complaints above, please add "AS" as done elsewhere in the statement. As a SQL newbie, this took me a while to figure out.
Also consider using clearer variable / alias names, especially given some aliases are used before the alias is even defined.

JPF/bad_buildings.sql
```diff
@@ -56,6 +80,16 @@ drop table if exists boro_lookup;
...
+drop table if exists hpd_buildings_temp;
+################
+
+create temporary table
+	hpd_buildings_temp
+select * from hpd_buildings where lifecycle = 'Building';
+alter table hpd_buildings_temp add primary key(buildingid);
 
 ################
 drop table if exists hs_complaint_count;
```
rdlester, line 95: Comment.
I'm gonna stop making these comments, hopefully you get the idea.

JPF/bad_buildings.sql
```diff
@@ -71,7 +105,7 @@ left join
 on
 	c.complaintid = cp.complaintid
 left join 
-	hpd_buildings b
+	hpd_buildings_temp b
 on 
 	c.buildingid = b.buildingid
 where
```
rdlester, line 111: What are these category ids? I don't know if SQL has inline comments, but it'd be nice to include the id name in a comment.

JPF/bad_buildings.sql
```diff
@@ -117,20 +151,22 @@ select
 	hb.lot as lot,
```
rdlester, line 151: Where does boroid in the above concat statement come from? Shouldn't it be hb.boroid?

JPF/bad_buildings.sql
```diff
@@ -139,48 +175,64 @@ drop table if exists hs_permit_counts;
...
 where 
 	work_type in ("OT","BL", "DM") 
 group by 
-	1,2;
-alter table hs_permit_counts add index(hs_dobp_address(255));
-alter table hs_permit_counts add index(bin_num);
+	1;
+#alter table hs_permit_counts add index(address(255));
+alter table hs_permit_counts add index(bin);
+#alter table hs_permit_counts add index(bbl);
```
rdlester, line 191: Just delete these lines if they are no longer needed. (You can always pull them from the git history)

JPF/bad_buildings.sql
```diff
@@ -139,48 +175,64 @@ drop table if exists hs_permit_counts;
...
 
 ################	
 drop table if exists permit_counts;
 ################
 create temporary table permit_counts
 select
```
rdlester, line 197: What's the difference b/w permit_counts and hs_permit_counts? Just the work_type filtering? This really needs comments.

JPF/bad_buildings.sql
```diff
@@ -97,7 +131,7 @@ left join
 on
 	c.complaintid = cp.complaintid
 left outer join 
-    hpd_buildings b
+    hpd_buildings_temp b
 on 
    c.buildingid = b.buildingid
 group by
```
rdlester, line 137: Using the full column name for group by here would be more readable, though it's not essential.

JPF/bad_buildings.sql
```diff
@@ -117,20 +151,22 @@ select
 	hb.lot as lot,
 	hb.bin as bin,
 	hb.zip as zip,
```
rdlester, line 153: Just aliasing columns to avoid typing out the table name feels like overkill, especially since you've already aliased the table name to shorthand.

JPF/bad_buildings.sql
```diff
@@ -139,48 +175,64 @@ drop table if exists hs_permit_counts;
...
+#alter table violation_counts add index(address(255));
 alter table violation_counts add index(bin);
+#alter table violation_counts add index(bbl);
+#DELETE FROM violation_counts where bbl like "% %";
+#delete from violation_counts where bbl like "%N%";
+#delete from violation_counts where bbl regexp '[A-Za-z]';
+
 
 ################
 drop table if exists classA_vio_cnt;
```
rdlester, line 238: I can't comment on the below but all the same comments apply.

The class?_vio_cnt tables are especially good places for stored procedures, I think.

JPF/bad_buildings.sql
```diff
@@ -348,9 +421,35 @@ drop table if exists bad_buildings;
 
 create table bad_buildings
```
rdlester, line 422: Hot dang this is a long SQL statement. Is there any way to break this down into easier-to-comprehend bits, or does SQL not really allow for that without compromising performance / requiring tons of temporary tables (like the aep yearly tables created directly above)?

JPF/bad_buildings.sql
```diff
@@ -455,10 +556,75 @@ left join
...
 ;
 
+## CREATE AND SET THE TARGET COLUMN
+ALTER TABLE bad_buildings ADD `target` INT  NULL  DEFAULT NULL  AFTER `managementprogram`;
+ALTER TABLE bad_buildings ADD `bb_target` INT  NULL  DEFAULT NULL  AFTER `target`;
+ALTER TABLE bad_buildings ADD `bb_score` FLOAT  NULL  DEFAULT NULL  AFTER `bb_target`;
+alter table bad_buildings add index(unitsres);
+alter table bad_buildings add index(pa_score);
+alter table bad_buildings add index(dob_violation_cnt);
+alter table bad_buildings add index(total_aep_yrs);
```
rdlester, line 570: These ALTER TABLE statements might be able to be combined into a single statement, with sub-statements separated by commas.

JPF/bad_buildings.sql
```diff
@@ -455,10 +556,75 @@ left join
...
 left join
 	aep_2016 a16 on (a16.full_addr) = (hbaddress)
+left join
+	hpd_buildings_temp hb on hb.buildingid = cc.buildingid
 ;
 
+## CREATE AND SET THE TARGET COLUMN
+ALTER TABLE bad_buildings ADD `target` INT  NULL  DEFAULT NULL  AFTER `managementprogram`;
+ALTER TABLE bad_buildings ADD `bb_target` INT  NULL  DEFAULT NULL  AFTER `target`;
+ALTER TABLE bad_buildings ADD `bb_score` FLOAT  NULL  DEFAULT NULL  AFTER `bb_target`;
```
rdlester, line 566: Dumb SQL question: does column order actually matter?

JPF/bad_buildings.sql
```diff
@@ -455,10 +556,75 @@ left join
...
+	bad_buildings 
+where 
+	pa_score is not null; 
+
+
+
+
+
+
+
```
rdlester, line 623: Super nit: get rid of blank lines.

JPF/bad_buildings.sql
```diff
@@ -477,26 +643,8 @@ alter table pluto_nyc drop index bbl;
 alter table tb_changes_summary drop index ucbbl;
 alter table hpd_aep_list drop index full_addr;
```
rdlester, line 644: Are these drop index statements really necessary? If SQL errors out if attempting to add an index on something already indexed, I guess it is in order for this SQL file to be re-runnable. Maybe it's worth extracting the initial table cleaning to a separate SQL file? It would also shorten this file a little and make it easier to take in.

rdlester (commented):
I just reviewed the SQL file. Mostly just nits that would drastically improve readability.

I can't really review the ipython notebooks since the format does not diff well at all. One potential solution to the diff problem is to use the nbconvert subcommand to convert the notebooks into regular python files - this would let us focus on the code and see only the code changes.

JPF/bad_buildings.sql
rdlester: I figured this out and thought I had deleted this comment but hadn't, sorry. I think the indexes are clear enough here, but a little confusing in the longer statements.

rdlester:
Just gonna go ahead and merge these in.

PR #32, opened 2016-12-13
HPD Complaints Script

chrishenry:
# Changelog

### New

* HPD Complaints script. [Chris Henry]

### Changes

* bring volumes in line with data-imports folder structure. [Chris Henry]

### Fix

* Send correct # of params to process_df. [Chris Henry]

### DIFF
--- a/data-imports/import-hpd-complaints.py
+++ b/data-imports/import-hpd-complaints.py
@@ -1,0 +1,153 @@
+#!/usr/bin/env python
+
+import argparse
+import os
+import os.path
+import sys
+import logging
+
+from utils import *
+
+mkdir_p(BASE_DIR)
+
+logging.basicConfig(format='[%(asctime)s] {%(pathname)s:%(lineno)d} %(levelname)s - %(message)s',
+    datefmt='%H:%M:%S',
+    stream=sys.stdout,
+    level=logging.INFO)
+log = logging.getLogger(__name__)
+
+"""
+HPD Buildings import
+"""
+
+hpd_complaints_KEY = 'hpd_complaints'
+
+cmp_dtype_dict = {
+    'ComplaintID':         'int64',
+    'BuildingID':          'int64',
+    'BoroughID':           'int64',
+    'Borough':            'object',
+    'HouseNumber':        'object',
+    'StreetName':         'object',
+    'Zip':               'float64',
+    'Block':               'int64',
+    'Lot':                 'int64',
+    'Apartment':          'object',
+    'CommunityBoard':      'int64',
+    'ReceivedDate':       'object',
+    'StatusID':            'int64',
+    'Status':             'object',
+    'StatusDate':         'object'
+}
+
+cmp_df_keep_cols = [
+    'ComplaintID',
+    'BuildingID',
+    'BoroughID',
+    'Borough',
+    'HouseNumber',
+    'StreetName',
+    'Zip',
+    'Block',
+    'Lot',
+    'Apartment',
+    'CommunityBoard',
+    'ReceivedDate',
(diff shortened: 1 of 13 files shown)


data-imports/import-hpd-complaints.py
```diff
@@ -0,0 +1,153 @@
...
+
+    cmp_description = "HPD Complaints"
+    cmp_input_csv_url = hpd_complaints_csv
+    cmp_sep_char = ","
+    cmp_pickle = os.path.join(hpd_complaints_dir, 'df_complaints.pkl')
+    cmp_table_name = 'hpd_complaints'
+    cmp_load_pickle = args.LOAD_PICKLE
+    cmp_save_pickle = args.SAVE_PICKLE
+    cmp_db_action = args.DB_ACTION
+    cmp_chunk_size = 5000
```
rdlester, line 93: Hmm, we've been using 2500 elsewhere, is it better to use larger chunks? Is there something about the HPD complaints that allow for a larger chunk size than the other scripts?

data-imports/import-hpd-complaints.py
```diff
@@ -0,0 +1,153 @@
...
+                cmp_dtype_dict,
+                cmp_load_pickle,
+                cmp_save_pickle,
+                cmp_pickle,
+                cmp_db_action,
+                cmp_truncate_columns,
+                cmp_date_time_columns,
+                cmp_chunk_size,
+                cmp_df_keep_cols
+               )
```
rdlester, line 109: Not for this PR but I'd like to drop a couple of these params and just pass the args struct directly, much less copy-pasta.

data-imports/utils.py
```diff
@@ -152,7 +152,7 @@ def hpd_csv2sql(description, input_csv_url, sep_char,
             log.info("Processing chunk {}".format(chunk_num))
 
             chunk_pickle_file = pickle_file.format(chunk_num)
-            df = process_df(df, log, description, save_pickle, chunk_pickle_file,
+            df = process_df(df, log, description, save_pickle, pickle_file,
```
rdlester, line 155: This is supposed to be chunk_pickle_file - when importing the CSV in chunks (only used for 311 right now), the user provides a pickle file name format, which gets formatted with the number (into chunk_pickle_file). Passing pickle_file will cause each chunk's pickle to overwrite that of the previous chunk.

I should probably pull this stuff out to a separate method so it's less confusing.

jpfreeley:
I'm not sure if there's anything significantly better or worse .. it's just
small amount of overhead .. less dense tables can support higher chunks. I
had basically started everything with 5K and if it didn't work, tried 2.5K
.

data-imports/import-hpd-complaints.py
chrishenry: Ahh, this was arbitrary, but it seemed to perform pretty well. `hpd_complaints` is a much narrower table than most of the others, which does allow for larger chunks.

data-imports/import-hpd-complaints.py
chrishenry: Yea, that'd be a great improvement. In the spirit of tracking changes we want to make, I'll move this to an issue.

https://github.com/chrishenry/deltanyc-heatseek/issues/33

data-imports/utils.py
chrishenry: Hmm, I wonder if there was a bad merge somewhere? Or if that param order is in a different PR? The changes I made here seem to correspond to what's in master now for `process_df`.

https://github.com/chrishenry/deltanyc-heatseek/blob/master/data-imports/utils.py#L192

PR #35, opened 2016-12-24
Get SQL cleanup working

rdlester:
Opening a pull request before all import scripts are updated to make sure everyone's okay with the general approach.

- Swtiched Mysql regex UDFs to a package I could actually get installed: https://github.com/mysqludf/lib_mysqludf_preg

- Added UDF installation to Dockerfile (testing this now).

- Abstracted shared cleanup commands into helper functions (data-imports/clean_utils.py - needs an underscore instead of a dash in the name in order to be imported in other scripts (?)). (tested)

- Updated import-dob-permits.py. (tested)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,13 +25,6 @@ RUN wget --quiet https://github.com/krallin/tini/releases/download/v0.10.0/tini 
     mv tini /usr/local/bin/tini && \
     chmod +x /usr/local/bin/tini
 
-# Install Mysql regex UDFs
-RUN wget --quiet https://github.com/mysqludf/lib_mysqludf_preg/archive/testing.zip && \
-    mkdir lib_mysqludf_preg && \
-    unzip testing.zip -d lib_mysqludf_preg && \
-    cd lib_mysqludf_preg && \
-    ./configure && make install && make installdb && make test;
-
 ENV SHELL /bin/bash
 ENV NB_USER root
 ENV LC_ALL en_US.UTF-8
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,12 +25,18 @@ RUN wget --quiet https://github.com/krallin/tini/releases/download/v0.10.0/tini 
     mv tini /usr/local/bin/tini && \
     chmod +x /usr/local/bin/tini
 
+# Install Mysql regex UDFs
+RUN wget --quiet https://github.com/mysqludf/lib_mysqludf_preg/archive/testing.zip && \
+    mkdir lib_mysqludf_preg && \
+    unzip testing.zip -d lib_mysqludf_preg && \
+    cd lib_mysqludf_preg && \
+    ./configure && make install && make installdb && make test;
+
 ENV SHELL /bin/bash
 ENV NB_USER root
 ENV LC_ALL en_US.UTF-8
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US.UTF-8
-
 
 # Upgrade pip
 RUN pip install --upgrade pip
--- a/data-imports/clean_utils.py
+++ b/data-imports/clean_utils.py
@@ -69,16 +69,19 @@ def make_index(table, column):
 
 def clean_bbl(table, boro, block, lot):
     return '''
-    SELECT concat(trim({boro}),trim(LPAD({block}, 5, '0')),
-            trim(LPAD({lot}, 4, '0'))) as bbl from hpd_buildings;
-    ALTER TABLE {table} CHANGE bbl bigint(13) NULL DEFAULT NULL;
-    ALTER TABLE {table} ADD INDEX bbl;
+    ALTER TABLE {table} ADD COLUMN bbl bigint(13) NULL DEFAULT NULL;
+    UPDATE {table} SET bbl =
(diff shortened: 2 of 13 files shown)


jpfreeley:
It looks good to me.

Dockerfile
```diff
@@ -25,13 +25,19 @@ RUN wget --quiet https://github.com/krallin/tini/releases/download/v0.10.0/tini
     mv tini /usr/local/bin/tini && \
     chmod +x /usr/local/bin/tini
 
+# Install Mysql regex UDFs
+RUN wget --quiet https://github.com/mysqludf/lib_mysqludf_preg/archive/testing.zip && \
+    mkdir lib_mysqludf_preg && \
+    unzip testing.zip -d lib_mysqludf_preg && \
+    cd lib_mysqludf_preg && \
+    ./configure && make install && make installdb && make test;
```
chrishenry: Hmm, I think this would need to be installed in the MySQL container.

Adding this will probably require us to build our own MySQL container. Or we could use [this](https://github.com/uqlibrary/docker-mysql-udf-preg)

data-imports/clean_utils.py
```diff
@@ -0,0 +1,40 @@
+def clean_addresses(table, column):
+    """ Returns a SQL statement cleaning an address column.
+        |table| and |column| are both strings.
+    """
+    # One line missing:
+    #    UPDATE {table} SET {column} = array_to_string(regexp_matches('(.*)(\d+)(?:TH|RD|ND|ST)( .+)'), '', {column} ) WHERE {column} ~ '.*(\d+)(?:TH|RD|ND|ST)( .+).*';
+    # No idea where regexp_matches is defined? Looks like it's a Postgresql command?
```
chrishenry, line 9: Yup, `regexp_matches` is a Postgres function, [documented here](https://www.postgresql.org/docs/9.1/static/functions-string.html)

Dockerfile
rdlester: Seems like the "this" will work fine. Tried adding it to the docker-compose.yml (pushed to this branch), but the DB failed to start up with a long list of errors, the most important bit of which I pasted below. Any ideas on what's wrong?

db_1   | 2016-12-24 16:51:34 7fa19625e720  InnoDB: Operating system error number 13 in a file operation.
db_1   | InnoDB: The error means mysqld does not have the access rights to
db_1   | InnoDB: the directory.
db_1   | 2016-12-24 16:51:34 13 [ERROR] InnoDB: Creating or opening ./ibdata1 failed!
db_1   | 2016-12-24 16:51:34 13 [ERROR] InnoDB: Could not open or create the system tablespace. If you tried to add new data files to the system tablespace, and it failed here, you should now edit innodb_data_file_path in my.cnf back to what it was, and remove the new ibdata files InnoDB created in this failed attempt. InnoDB only wrote those files full of zeros, but did not yet use them in any way. But be careful: do not remove old data files which contain your precious data!

chrishenry: Ouch, those are file permission errors. As in, MySQL does not have permission to write to your datadir. This was pretty common with older versions of the MySQL container, and solved in a workable way earlier this year.

Looking at the [build history for the docker-mysql-udf-preg container](docker-mysql-udf-preg), the last build was a year ago, probably before the MySQL container worked out the kinks.

Options to move forward are, presented on a range of correct -> fast;

1) Poke the owner, see if he could push a new build. 
2) Fork the repo, and do it ourselves. (I'd be happy to do this)
3) Create a local Dockerfile for MySQL, copy-pasta the pertinent line, and have `docker-compose` manage the build for us.

rdlester:
My vote is number 2, and maybe send a pull request back to the original
repo. But, looking through the Dockerfile, I didn't see anything linked to
a particular version of the mysql container, so I'm not sure that's the
issue.

rdlester:
Pushed some new updates:

- The code obtaining the cursor / connection was wrong; fixed and pulled out into a helper.

- Updated 311 import. Will keep updating later.

rdlester:
Looked more into the Docker stuff. One issue is that I wasn't pointing at the right place for the container (needed to use a git:// url, using the <name>/<repo> format points at docker.io?).

I forked the repo (https://github.com/rdlester/docker-mysql-udf-preg) and updated the install to build from HEAD but am still getting the same errors.

data-imports/clean_utils.py
```diff
@@ -0,0 +1,91 @@
...
+
+def make_index(table, column):
+    """ Returns an SQL statement that makes an index on the given column.
+    """
+    return "ALTER TABLE {} ADD INDEX {};".format(table, column)
+
+def clean_bbl(table, boro, block, lot):
+    return '''
+    SELECT concat(trim({boro}),trim(LPAD({block}, 5, '0')),
+            trim(LPAD({lot}, 4, '0'))) as bbl from hpd_buildings;
```
rdlester: SELECT ... AS also appears to be Postgresql...

rdlester:
The bbl SQL code has been modified to something using standard SQL.

I've tested all the import scripts and they all seem to work. I haven't looked too closely at the output (I expect the address cleaning is still not good enough to link 100% of things together), and I do know there's a few cases that need additional cleaning (one table has additional boro codes "0" and "undefined"), but this is a good start.

Please take another look.

chrishenry:
👍

Issue #34, opened 2016-12-19
Add Mysql regex UDF to Docker setup


rdlester:
Two parts here:

1. Set up Docker instance to install https://github.com/mysqludf/lib_mysqludf_preg. There's a one-line command line install in the README; should just work out of the box on our instances (as PCRE is already installed).

2. Convert existing usage of regex_replace to preg_replace. The regex format is "perl" style (basically need to surround the regex with slashes); the column name goes last instead of first.

chrishenry:
Resolved in https://github.com/chrishenry/deltanyc-heatseek/pull/38

PR #38, opened 2016-12-29
Create UDFs in MySQL container

chrishenry:
A few notes;

* The container provided at [Dockerhub](https://hub.docker.com/r/uqlibrary/docker-mysql-udf-preg/) hasn't been built for over a year. This means the Docker image available was built using a very old upstream container, hence most of the problems we're seeing. 
* This PR automates the process of installing the required binaries into the db container, and bringing up a quick helper container (`db-udf`) to run the necessary `CREATE FUNCTION` commands on start of docker-compose.

# Changelog

### New

* Update README with preg install notes. [Chris Henry]

* Create a helper container that will run the CREATE FUNCTION cmds for the preg* functions. [Chris Henry]

* Create new docker file for installing preg udfs. [Chris Henry]

### DIFF
--- a/Dockerfile.mysql
+++ b/Dockerfile.mysql
@@ -1,0 +1,12 @@
+FROM mysql:latest
+
+RUN apt-get update && apt-get install -y wget ca-certificates build-essential libmysqlclient-dev libpcre3-dev && \
+      wget -O lib_mysqludf_preg-1.2-rc2.tar.gz https://github.com/mysqludf/lib_mysqludf_preg/archive/lib_mysqludf_preg-1.2-rc2.tar.gz && \
+      tar xzf lib_mysqludf_preg-1.2-rc2.tar.gz && \
+      cd lib_mysqludf_preg-lib_mysqludf_preg-1.2-rc2 && \
+      ./configure && \
+      make install && \
+      apt-get remove -y wget ca-certificates build-essential libmysqlclient-dev libpcre3-dev && \
+      cat installdb.sql >> /tmp/mysql-first-time.sql && \
+      cd .. && rm -rf lib_mysqludf_preg-lib_mysqludf_preg-1.2-rc2 && \
+      apt-get clean && apt-get purge
--- a/Dockerfile.mysql
+++ b/Dockerfile.mysql
@@ -7,6 +7,5 @@ RUN apt-get update && apt-get install -y wget ca-certificates build-essential li
       ./configure && \
       make install && \
       apt-get remove -y wget ca-certificates build-essential libmysqlclient-dev libpcre3-dev && \
-      cat installdb.sql >> /tmp/mysql-first-time.sql && \
       cd .. && rm -rf lib_mysqludf_preg-lib_mysqludf_preg-1.2-rc2 && \
       apt-get clean && apt-get purge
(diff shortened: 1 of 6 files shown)


Dockerfile.mysql
```diff
@@ -0,0 +1,12 @@
+FROM mysql:latest
+
+RUN apt-get update && apt-get install -y wget ca-certificates build-essential libmysqlclient-dev libpcre3-dev && \
+      wget -O lib_mysqludf_preg-1.2-rc2.tar.gz https://github.com/mysqludf/lib_mysqludf_preg/archive/lib_mysqludf_preg-1.2-rc2.tar.gz && \
+      tar xzf lib_mysqludf_preg-1.2-rc2.tar.gz && \
+      cd lib_mysqludf_preg-lib_mysqludf_preg-1.2-rc2 && \
+      ./configure && \
+      make install && \
+      apt-get remove -y wget ca-certificates build-essential libmysqlclient-dev libpcre3-dev && \
+      cat installdb.sql >> /tmp/mysql-first-time.sql && \
```
rdlester: Do we still need this line? Doesn't seem like that first-time sql is ever actually run (hence the need for mysql-udf).

README.md
```diff
@@ -60,3 +58,16 @@ on why you need to use `127.0.0.1`.
...
+The cleanup utils make extensive use of [User Defined preg functions](https://github.com/mysqludf/lib_mysqludf_preg).
+These aren't available in the general MySQL docker container, and the container
+provided by the [uqlibrary](https://github.com/uqlibrary/docker-mysql-udf-preg)
+hasn't been built in a while, so it doesn't have recent fixes to the upstream
+MySQL container.
+
+One challenge here is that the `CREATE FUNCTION` statements need to be called by
+a client, as MySQL UDFs are registered in the `mysql` database. This was solved
+by creating a quick utility container (`db-udf`) that will come up with
+docker-compose, pull the relevant `CREATE` commands, run them, and exit.
```
rdlester, line 72: So, just to make sure - the db-udf container only needs to run once? (So if I just want to boot up the DB container, I should `docker compose up db db-udf`?)

rdlester:
👍

README.md
```diff
@@ -30,17 +30,15 @@ app.
```
rdlester: Should we add a note to "option 1" mentioning these UDFs should be installed as well? (Just a link to the github repo is likely enough)

README.md
chrishenry: Yep. 

Sadly, on subsequent runs of plain old `docker-compose up`, the `db-udf` container will exit with an error. MySQL will error out upon trying to create the preg functions again. It'll look like this;

```
db-udf_1  | ERROR 1125 (HY000) at line 3: Function 'lib_mysqludf_preg_info' already exists
```

For now, I think it's OK, but I've created an issue for it to be cleaned up.
https://github.com/chrishenry/deltanyc-heatseek/issues/39

README.md
chrishenry: Good call. Will do.

chrishenry:
Resolves https://github.com/chrishenry/deltanyc-heatseek/issues/34

Issue #33, opened 2016-12-16
Pass args struct directly into hpd_csv2sql


chrishenry:
`utils.hpd_csv2sql` has a ton of params, and implementing a new instance typically results in some copy-pasta.

To simplify arguments, pass the args struct.

Additionally, consider changing more parameters to keywords, which lends some clarity.

PR #45, opened 2017-01-06
Cleanup

rdlester:
Refactoring stuff to be simpler.

- hpd2csv takes the args struct directly, reducing the number of params.
- utils now performs all arg parsing.
- Simplified format of import scripts so data is listed in files in the same order.
- Various sensical pylint and pyflakes fixes.

Untested as of now but will test. Don't merge yet tho just comment


rdlester:
Added a "test mode" for the import scripts. It's untested tho

rdlester:
Fixes #33

chrishenry:
LGTM

PR #57, opened 2017-01-12
data-imports only requirements.txt

rdlester:
I was running into errors spinning up a new virtualenv from the root requirements.txt, so I installed only the stuff I seemed to need for data-imports and pip freeze-d it into a new requirements.txt. This stuff should be sufficient for running these scripts.


chrishenry:
LGTM, and non-breaking

Issue #53, opened 2017-01-11
error in import-aep.py


LKWLaLa:
screen shot:

![image](https://cloud.githubusercontent.com/assets/15372198/21830850/b6b1ba5a-d76d-11e6-8af4-de1cbf809d85.png)

rdlester:
Your working directory needs to be the root of this repo when running this script - data/AEP_list.csv is a relative directory.

We should change this to build the absolute path of the data file, which can be determined given the absolute path of the script.

PR #73, opened 2017-01-27
AEP Import

chrishenry:
Fixes https://github.com/chrishenry/deltanyc-heatseek/issues/53

### DIFF
--- a/data-imports/import-aep.py
+++ b/data-imports/import-aep.py
@@ -33,7 +33,7 @@ def main():
 
 def import_csv(args):
     log.info('Importing pre-created AEP csv...')
-    data_file = '/root/data/AEP_LIST.csv'
+    data_file = os.path.abspath(os.path.join(os.path.dirname(os.path.realpath(__file__)), '../data/AEP_LIST.csv'))
     df = pd.read_csv(data_file)
 
     conn = connect(args.TEST_MODE)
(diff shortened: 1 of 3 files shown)


data-imports/import-aep.py
```diff
@@ -33,10 +33,10 @@ def main():
 
 def import_csv(args):
     log.info('Importing pre-created AEP csv...')
-    data_file = 'data/AEP_LIST.csv'
+    data_file = '/root/data/AEP_LIST.csv'
```
rdlester: Adding '/root' means it can't be run outside of Docker. We could just decide these scripts can't be run outside of docker, but I'd like to keep the option available. What's the working directory the script is running from in Docker, and how can we map things so the script can be run from the root of the repo?

(Either way the import scripts could use a README, I can work on that)

data-imports/import-aep.py
```diff
@@ -33,10 +33,10 @@ def main():
 
 def import_csv(args):
     log.info('Importing pre-created AEP csv...')
-    data_file = 'data/AEP_LIST.csv'
+    data_file = os.path.abspath(os.path.join(os.path.dirname(os.path.realpath(__file__)), '../data/AEP_LIST.csv'))
```
chrishenry, line 36: OK, this incantation should build out an absolute path (allowing the script to be run from anywhere) to the AEP file.

As long as the directory structure is consistent, this should work in both Docker and elsewhere.

rdlester: Looks good to me.

rdlester (approved):

PR #74, opened 2017-01-27
Fix import-pluto

chrishenry:
Fixes https://github.com/chrishenry/deltanyc-heatseek/issues/68

# Changelog

### New

* Strip leading and trailing whitespace when running clean_address. [Chris Henry]

* Add a simple bashrc. [Chris Henry]

### Fix

* Use the correct variable when building csv path. [Chris Henry]

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,6 +39,7 @@ COPY docker/start.sh /usr/local/bin/
 COPY docker/start-notebook.sh /usr/local/bin/
 COPY docker/start-singleuser.sh /usr/local/bin/
 COPY docker/jupyter_notebook_config.py /root/.jupyter/
+COPY docker/.bashrc /root/.bashrc
 COPY requirements.txt /root/requirements.txt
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -39,6 +39,7 @@ COPY docker/start.sh /usr/local/bin/
 COPY docker/start-notebook.sh /usr/local/bin/
 COPY docker/start-singleuser.sh /usr/local/bin/
 COPY docker/jupyter_notebook_config.py /root/.jupyter/
+COPY docker/.bashrc /root/.bashrc
```
rdlester, line 42: Was this necessary to get import-pluto working?

rdlester (approved):

Dockerfile
chrishenry: Nope, it's just nice to have some simple bash aliases.

Virtually no Docker these containers ship with any bash customization (some don't even ship with a shell). I really just got sick of not being able to type `ll`. 😄

Issue #62, opened 2017-01-14
SSL error when running rake db_connector:three11


LKWLaLa:
```
**************
BX

**************
MN

**************
BK

rake aborted!
OpenSSL::SSL::SSLError: SSL_connect SYSCALL returned=5 errno=0 state=unknown state
/usr/local/bundle/gems/httparty-0.13.7/lib/httparty/request.rb:117:in `perform'
/usr/local/bundle/gems/httparty-0.13.7/lib/httparty.rb:545:in `perform_request'
/usr/local/bundle/gems/httparty-0.13.7/lib/httparty.rb:476:in `get'
/usr/local/bundle/gems/httparty-0.13.7/lib/httparty.rb:583:in `get'
/angular-rails-app/lib/tasks/db_connector.rake:28:in `nyc_geocode'
/angular-rails-app/lib/tasks/db_connector.rake:394:in `block (3 levels) in <top (required)>'
/angular-rails-app/lib/tasks/db_connector.rake:382:in `each'
/angular-rails-app/lib/tasks/db_connector.rake:382:in `block (2 levels) in <top (required)>'
/usr/local/bundle/gems/rake-12.0.0/exe/rake:27:in `<top (required)>'
Tasks: TOP => db_connector:three11
```

Was previously getting a Net::OpenTimeout error : 
```**************
QN

**************
BX

**************
BK

**************
MN

rake aborted!
Net::OpenTimeout: Net::OpenTimeout
/usr/local/bundle/gems/httparty-0.13.7/lib/httparty/request.rb:117:in `perform'
/usr/local/bundle/gems/httparty-0.13.7/lib/httparty.rb:545:in `perform_request'
/usr/local/bundle/gems/httparty-0.13.7/lib/httparty.rb:476:in `get'
/usr/local/bundle/gems/httparty-0.13.7/lib/httparty.rb:583:in `get'
/angular-rails-app/lib/tasks/db_connector.rake:28:in `nyc_geocode'
/angular-rails-app/lib/tasks/db_connector.rake:394:in `block (3 levels) in <top (required)>'
/angular-rails-app/lib/tasks/db_connector.rake:382:in `each'
/angular-rails-app/lib/tasks/db_connector.rake:382:in `block (2 levels) in <top (required)>'
/usr/local/bundle/gems/rake-12.0.0/exe/rake:27:in `<top (required)>'
Tasks: TOP => db_connector:three11
```

LKWLaLa:
[...]

PR #76, opened 2017-02-02
311 connector fixes

chrishenry:
# Changelog

### New

* Add additional 311 and dob_violation categories as per Noelle. [Chris Henry]

### Changes

* Simplify error handling. [Chris Henry]

* Add_indexes now takes an array of arrays, instead of hash. [Chris Henry]

Should help with https://github.com/chrishenry/deltanyc-heatseek/issues/62

### DIFF
--- a/angular-rails-app/lib/tasks/db_connector.rake
+++ b/angular-rails-app/lib/tasks/db_connector.rake
@@ -388,30 +388,22 @@ namespace :db_connector do
 
     three11_categories.each do |category|
 
-      sql = "SELECT * FROM call_311 WHERE incident_address IS NOT NULL AND borough != 'unspecified' AND complaint_type = " + category + ";"
-      puts sql
+      sql = "SELECT * FROM call_311 WHERE incident_address IS NOT NULL AND borough != 'unspecified' AND complaint_type = \"" + category + "\";"
       three11_results = conn.exec_query(sql).to_hash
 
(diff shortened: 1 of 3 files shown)


angular-rails-app/lib/tasks/db_connector.rake
```diff
@@ -299,6 +302,8 @@ namespace :db_connector do
       "LBLVIO-LOW PRESSURE BOILER",
       "LL1081-LOCAL LAW 10/81 - ELEVATOR",
       "LL6291-LOCAL LAW 62/91 - BOILERS",
+      "C-CONSTRUCTION",
+      "B-BOILER",
     ]
```
rdlester, line 307: Something that was useful for me in the data-import scripts was to pull important lists of data or column names out of the functions and at the top of the file. Those were the interesting bits, the import function was mostly boilerplate. That's not really true here, but it'd be nice to keep these lists of categories as easy to find as possible. You've already got a few vars up at the top, maybe it's worth adding the dob_violation and 311 types up to the top, or at least rename this variable to make it more obvious?

rdlester (approved):

angular-rails-app/lib/tasks/db_connector.rake
chrishenry: Yup, good call. Since we're working within Rails, I moved this list into the model.

PR #78, opened 2017-02-05
Write rspec tests for Rails property model and properties controller, install jasmine/karma and dependencies

LKWLaLa:
### Rails ###

- Add `PropertiesController` tests in rspec for `#index` and `#show` methods.  `#query` method tests are pending

- Add `Property` model tests in rspec

- Install FactoryGirl and other dependencies for rspec

- Create Factory for testing properties

### Angular ###

- Install jasmine, karma, and angular mocks for testing

### ToDo ###

- [ ]  Figure out how to set up karma with rails sprockets, to allow inclusion of dependency injections installed via the Gemfile/bundler

- [ ] Add rspec test for `PropertiesController` `#query` method  (will this method change?)

- [ ] Add rspec tests for `OwnersController` and other models

### QA steps ###

- [ ] Rebuild docker web container to account for updates to Gemfile

- [ ] Run `rspec` command from within web container to run Rails tests


angular-rails-app/app/assets/javascripts/property/property.html
```diff
@@ -1,8 +1,7 @@
 <section style="min-height: 65px;margin-top:105px;">
   <div class="container">
     <p id="breadcrumb">
-      <a class="breadcrumb_top" ui-sref="home">Home /</a> Building /
-      {{vm.data.street_address | titleCase}}, {{vm.data.city}} {{vm.data.state}}
+      <a class="breadcrumb_top" ui-sref="home">Home /</a> Building
```
chrishenry, line 4: Hmm, wasn't the address in the design?

angular-rails-app/db/migrate/20161217230548_add_work_type_to_dob_permit.rb
```diff
@@ -1,5 +0,0 @@
-class AddWorkTypeToDobPermit < ActiveRecord::Migration
-  def change
-    add_column :dob_permits, :work_type, :string
-  end
-end
```
chrishenry, line 5: Yep, sorry, I don't know how the duplicate got in here...

angular-rails-app/spec/controllers/properties_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    it "returns a successful 200 response" do
+       get :index, format: :json
+      expect(response).to be_success
+    end
+
+    it "returns all properties" do
+      FactoryGirl.create_list(:property, 5)
+      get :index, format: :json
+      parsed_response = JSON.parse(response.body)
+      expect(parsed_response.length).to eq(5)
```
chrishenry: Hmm, is this supposed to test that 5 properties will come back in json format? If so, you should also test that paging works. should be pretty easy to by using `FactoryGirl` to create 31 properties, and then ensure only 30 are returned.

chrishenry:
One question we should discuss briefly; whether or not to check in `bower_components`. [Here's a quick list of pros/cons.](https://addyosmani.com/blog/checking-in-front-end-dependencies/)

To add to them; checking them in can make diffs very, very hard to read. This PR is an example of that unfortunate fact.

My take is that for this particular project, it'll be simpler to have our dependencies checked in to avoid deploy problems and the inevitable build problems that result from the additional step. 

What do you guys think?

angular-rails-app/app/assets/javascripts/property/property.html
LKWLaLa: Yes - it was originally.  However, at our last meeting Andrew told me to get rid of it.  So, I actually made this change with him in person.

angular-rails-app/spec/controllers/properties_controller_spec.rb
LKWLaLa: Makes sense!  Sure thing - I'll make the change.

angular-rails-app/app/assets/javascripts/property/property.html
chrishenry: Okie dokie. It is 30px right below, at font-size 30px. ;)

LKWLaLa: Yeah.... the CSS needs to be cleaned up on this page.  :)  What is 30px though?  I'm not sure which part you are referring to..?

angular-rails-app/db/migrate/20161217230548_add_work_type_to_dob_permit.rb
LKWLaLa: no prob! :)

LKWLaLa:
@chrishenry 

I updated the property controller spec to check pagination as suggested, and seems to be working!  

Regarding checking in bower components - admittedly, I am not versed in DevOps, but it does seem (from reading the post you linked) that checking them in might be the better option for our particular situation.  The annoyances might be offset by the fact that it makes our lives (and someone else's) easier once we get to deployment.

angular-rails-app/app/assets/javascripts/property/property.html
chrishenry: Ah, I was saying that the address is just below, so there's really no need for it in the breadcrumbs. 😄

LKWLaLa: Oh!  Ha - yeah, hard to miss it.  :)

LKWLaLa:
@chrishenry  I created issues for my ToDo items above, so we could merge this as long as it looks ok to you, and we decide what we want to do about bower_components.  I'm still a bit stuck on the Karma configuration, but if other people have access, perhaps someone else might be interested in playing with / cleaning up what I got started.  We could add an issue to set up CI as well.  Not sure if that would be something to do now or after more tests are written.

chrishenry:
Merging, and just noting that committing assets is our best bet for now.

PR #80, opened 2017-02-11
Fix csv chunking and memory issues

rdlester:
For some reason, my new computer seems to harshly enforce memory constraints when running Docker. Both import-pluto and import-dob-violations died before completion, with nothing but a "Killed" printed to the terminal for explanation. Reducing the memory used by both scripts (by using csv import chunking and rewriting the csv concatenation routine) allowed them to run successfully.

I also noticed that csv chunking was totally broken; everyone probably needs to re-run import-311.

### DIFF
--- a/data-imports/import-pluto.py
+++ b/data-imports/import-pluto.py
@@ -239,6 +239,7 @@ def import_csv(args):
     chunk_size = 5000
     max_column_size = 255
     date_format = "%m/%d/%Y"
+    csv_chunk_size = 250000
 
     hpd_csv2sql(
                 description,
@@ -253,6 +254,7 @@ def import_csv(args):
                 chunk_size,
                 max_column_size,
                 date_format=date_format,
+                csv_chunk_size=csv_chunk_size
                )
 
 
--- a/data-imports/import-dob-violations.py
+++ b/data-imports/import-dob-violations.py
@@ -97,6 +97,7 @@ def import_csv(args):
     chunk_size = 5000
     max_col_len = 255
     date_format = "%Y%m%d"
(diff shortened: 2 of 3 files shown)


data-imports/import-pluto.py
```diff
@@ -229,7 +229,7 @@ def import_csv(args):
 
         sys.stdout.write("\rConcatenating Pluto boro csvs....")
         pluto_csv_files_dir = os.path.join(csv_dir, "BORO_zip_files_csv")
-        pandas_concat_csv(pluto_csv_files_dir, csv_file)
+        simple_concat_csv(pluto_csv_files_dir, csv_file)
```
chrishenry: Hmm, simply concat'ing the pluto files together resulted in some oddities. If I recall correctly, the columns don't line up across pluto files, and using pandas to concat resolved it. @jpfreeley and @mcleanalec, does that sound right?

After you re-imported, did you see any issues?

chrishenry: @jpfreeley @mcleanalec

rdlester: The issue with Pluto, as mentioned to me by Ziggy, is that the schema changes on an almost yearly basis. A quick look over the CSVs didn't reveal any differences but I didn't look super closely.

Part of the reason I was running into these memory issues was I had Docker set to cap out at 2GB; I ran into a similar issue with the web stuff and got around it by lifting the cap to 4GB.

I'll undo the pluto CSV concat change for now. I think the real best solution is to actually import each year individually.

rdlester: Done.

PR #87, opened 2017-02-13
Rename Geoclient env vars to match official API wording

rdlester:
I set myself up with a Geoclient API key today. The mismatch b/w wording on the official API page and the .env.example vars was a little confusing, so I've changed our vars to match the official names.


chrishenry:
:+1:

PR #95, opened 2017-02-21
ffaker import

rdlester:
Guess this doesn't matter any more, now that FFaker use is commented out, but I had to add a require statement to make db:setup work. Also stripped unnecessary whitespace while I was around.

### DIFF
--- a/angular-rails-app/db/seeds.rb
+++ b/angular-rails-app/db/seeds.rb
@@ -1,4 +1,6 @@
 =begin
+require 'ffaker'
+
 50.times do
      Property.create(
         street_address: FFaker::AddressUS.street_address,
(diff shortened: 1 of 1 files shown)


LKWLaLa:
@rdlester  Thanks!  Actually, feel free to use/uncomment FFaker.  I have been seeding my local database on an as needed basis, so I've changed the FFaker seed file around a few times.  If the current state is commented out, it is only because I did not want to re-run certain portions of it.  My mistake that it got pushed that way.  I've also taken for granted that I was the only one using certain things up to this point - so please feel free to give me a poke if anything is off.

rdlester:
@LKWLaLa I didn't intend to use the seed file, it just ran as part of db:setup...? I'll ping you offline on how to actually use the seed file.

LKWLaLa:
@rdlester Ah...  gotcha.  I've never tried db:setup, as I tend to run db:create, db:migrate, and db:seed separately.

PR #96, opened 2017-02-21
Cleanup and fixes

chrishenry:
# Changelog

### New

* Add more ignores to ensure data isn't actually committed. [Chris Henry]

### Changes

* Update owners rake task. [Chris Henry]

* Document that boros are to be passed around as 2 letter codes. [Chris Henry]

* Major cleanup of properties connector. [Chris Henry]

* Remove non-rails tables. [Chris Henry]

* Remove litigations from notebook (they were previously migrated to an import script) [Chris Henry]

These changes should result in some clearer output, as well as cleaner data.

For example;

```bash
root@6e53cda7c1bb:/angular-rails-app# rake db_connector:properties
Copying from hpd_buildings...done
Copying from pluto_nyc...done
Found 10370 addresses without registrationid...fixing.
Finished importing properties
```

Should also solve https://github.com/chrishenry/deltanyc-heatseek/issues/104


angular-rails-app/lib/tasks/db_connector.rake
```diff
@@ -49,32 +49,39 @@ namespace :db_connector do
...
-      sql = "UPDATE hpd_buildings SET boro = '#{key}' WHERE boro = '#{value}';"
-      conn.execute(sql)
-    end
-
-    # Load addresses from HPD. This will catch additional addresses in hpd that are *not* in pluto
+    # Load addresses from HPD.
+    print "Copying from hpd_buildings..."
     sql = "INSERT IGNORE INTO r_properties (street_address,city,state,zipcode,hpd_registration_id,borough,block,lot,created_at,updated_at)
-    SELECT TRIM(streetname), 'New York', 'New York', zip, registrationid, boro, block, lot, NOW(), NOW() FROM hpd_buildings;"
+    SELECT TRIM(streetname), 'New York', 'New York', zip, registrationid, boro, block, lot, NOW(), NOW() FROM hpd_buildings WHERE streetname != '' AND streetname IS NOT NULL AND registrationid != 0;"
```
rdlester: "New York" does not seem like it would always be the right city, given boroughs are used for postal addresses...?

angular-rails-app/db/schema.rb
```diff
@@ -13,286 +13,8 @@
...
-    t.string   "zonemap",       limit: 255
-    t.string   "zmcode",        limit: 255
-    t.string   "sanborn",       limit: 255
-    t.float    "taxmap",        limit: 24
-    t.string   "edesignum",     limit: 255
-    t.float    "appbbl",        limit: 24
-    t.datetime "appdate"
-    t.integer  "plutomapid",    limit: 8
-  end
-
```
rdlester: uh what happened to the schemas?

rdlester (approved):

angular-rails-app/db/schema.rb
chrishenry: This was cleanup of tables that the Rails application doesn't have to know about.

angular-rails-app/lib/tasks/db_connector.rake
chrishenry: Riight. Would borough be a better choice for that field then?

rdlester: I think so, yes.

PR #97, opened 2017-02-21
implement Owners async search and alert box if not found + some general cleanup of existing CSS

### LINKED ISSUE #58: owner search on home page not yet functional
Needs to query Rails tables in the same way the Properties search does, as well as implementation of autocomplete.   Can probably implement the autocomplete with the Typeahead directive here:  https://angular-ui.github.io/bootstrap/   (Depending on the amount of data in the Owners table. If too much data is loaded into the front end at once, we might need a different solution.) The `ui.bootstrap` module is already loaded as a dependency in `app.js`, because I'm also using it for alerts/error messages.

Also create alert box if owner is not found in database (in the same way as is implemented for the properties search.  See `home.controller.js` for details.)

### LINKED ISSUE #84: Clean up CSS in general
Clean up CSS (especially the `custom.css` file and hard coded css in the `property.html` file) to follow best practices.  Consolidate and create grouping for readability and extensibility.

LKWLaLa:
# Changelog #

### New ###

- Implement Typeahead directive from Angular UI Bootstrap to create Owners dropdown menu

- Add conditional return of queried owners in #index method of OwnersController in Rails.  (This query will run when a `$viewValue` is submitted via the Owners search input box.

- Add getOwners() method in `home.service.js`

- Add getOwners() method in `home.controller.js` to return correct data options to the Typeahead directive

- Add notes to `home.controller.js` to organize page into Properties, Owners, and Alerts sections (for ease of reading)

- Add alert box with message under the owners search bar, if a particular owner is not found in the database

- General refactoring / cleanup of existing CSS

After review, this closes #58 and closes #84

### DIFF
--- a/angular-rails-app/app/controllers/owners_controller.rb
+++ b/angular-rails-app/app/controllers/owners_controller.rb
@@ -3,9 +3,14 @@ class OwnersController < ApplicationController
    before_action :set_owner, only: [:show]
 
   def index
-    @owners = Owner.all
+    if params[:name]
+       @owners = Owner.where("name LIKE (?)", "%#{params[:name]}%")
+    else
+      @owners = Owner.all
+    end
     render json: @owners, status: 200
   end
+
 
   def show
     render json: @owner, status: 200
--- a/angular-rails-app/app/controllers/owners_controller.rb
+++ b/angular-rails-app/app/controllers/owners_controller.rb
@@ -4,7 +4,7 @@ class OwnersController < ApplicationController
 
   def index
     if params[:name]
-       @owners = Owner.where("name LIKE (?)", "%#{params[:name]}%")
+       @owners = Owner.where("name LIKE (?)", "%#{params[:name]}%").limit(6)
     else
       @owners = Owner.all
     end
--- a/angular-rails-app/app/assets/javascripts/home/contact.html
+++ b/angular-rails-app/app/assets/javascripts/home/contact.html
@@ -1,9 +1,9 @@
 <div class="container" style="background: rgb(127,128,129); height: 100vh; width:100vw; margin-top: 0;">
 
-  <h2 id="resource_text" style="margin-top: 150px;">Contact</h2>
+  <h2 id="resource-text" style="margin-top: 150px;">Contact</h2>
   <hr class="featurette-divider resource">
 
-  <div id="resource_text">
+  <div id="resource-text">
     <a href="http://heatseek.org/" target="_blank">Heatseek.org</a><br>
     <a href="mailto:info@heatseek.org">info@heatseek.org</a>
   </div>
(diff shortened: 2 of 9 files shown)


angular-rails-app/app/controllers/owners_controller.rb
```diff
@@ -3,10 +3,15 @@ class OwnersController < ApplicationController
    before_action :set_owner, only: [:show]
 
   def index
-    @owners = Owner.all
+    if params[:name]
+       @owners = Owner.where("name LIKE (?)", "%#{params[:name]}%").limit(6)
```
chrishenry, line 7: Nice!

You should add an `OR` for `corporation_name` as well, since that's a valid search.

chrishenry (changes requested):

angular-rails-app/app/controllers/owners_controller.rb
LKWLaLa: Oh, good point!  Ok, yeah - I'll add that.  Thanks!

LKWLaLa: @chrishenry Actually, this is going to make things much more complicated.  Because I can't only do this change on the back end.  I would have to display either names or corporation names in the drop down box the front end, depending on what what matched - so we'd have to keep track of that somehow...   I think we should either limit the search to name only at this point in time, or redesign our schema altogether so that we have a corporation owner type that is separate from an individual owner type or employee of a corporation.  Thoughts?

LKWLaLa:
Did some additional CSS cleaning last night, so this PR should also resolve issue #84.

angular-rails-app/app/controllers/owners_controller.rb
LKWLaLa: PS - I think my preference would be to merge this as is for the moment, and create a separate issue of addressing the corporation name question.  But let me know your thoughts.

LKWLaLa: @chrishenry  Actually, I'm not sure why this didn't occur to me before, but perhaps a more simple answer for now is just to concat `name` + `corporation_name` and display both in the dropdown box.  I think I'm going to try this...

LKWLaLa: @chrishenry:  Ok, I've added the `OR` statement in the Rails query, so that we include matches from both the `name` and `corporation_name`columns.  In addition, I am concatenating `name` + `corporation_name` to display both together in the front end dropdown box.  Works pretty well! However, per my recent message on Slack, `corporation_name` is currently type: `integer`.  I've written a new migration to change this to a string.  Is that the intention?  (Just waiting on confirmation of `string` vs `integer` before adding the changes to this PR.)

chrishenry: yep, just saw the slack. I've created an [issue](https://github.com/chrishenry/deltanyc-heatseek/issues/104).

chrishenry (approved):

PR #103, opened 2017-02-23
--force-yes to db

rdlester:
This makes the db container install work but I'm not sure if it's right.


chrishenry:
Generally, it's not the best idea to blindly answer yes. 

However, since this isn't particularly sensitive work, I think it's fine.

PR #105, opened 2017-02-26
Allow query by owner's corporation_name as well as name

LKWLaLa:
# Changelog #

### New ###

- Display `corporation_name` in Owners search dropdown in addition to `name`

- Query both `name` and `corporation_name` in Rails OwnersController `#index` action


chrishenry:
👍

chrishenry (commented):
👍

PR #109, opened 2017-03-04
Rails test cleanup and Travis integration

### LINKED ISSUE #100: implement Travis CI

chrishenry:
# Changelog

### New

* Add a travis.yml file. [Chris Henry]

### Changes

* When property id not found, show a 404. [Chris Henry]

* Undo some bad debug work. [Chris Henry]

* Some fixes to get tests working. [Chris Henry]

Resolves https://github.com/chrishenry/deltanyc-heatseek/issues/100


docker-compose.yml
```diff
@@ -29,8 +29,8 @@ services:
       - ./angular-rails-app/:/angular-rails-app/
     ports:
       - "3000:3000"
-    depends_on:
-      - db
+    # depends_on:
```
rdlester: This doesn't seem right...

angular-rails-app/Gemfile
```diff
@@ -48,11 +48,12 @@ gem 'httparty', '~> 0.13.6'
...
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
+  gem 'sqlite3'
   gem 'byebug'
   gem 'thor', '0.19.1'
   gem 'rspec-rails', '~> 3.5'
   gem 'ffaker'
-  gem 'factory_girl_rails'
+  gem 'factory_girl_rails', :require => false
 end
```
rdlester, line 57: The Travis CI docs recommend adding "Rake" to the gem list somewhere. We're running a script instead of Rake, but still seems like good form.

.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+rvm:
+  - 2.3.3
+
+services:
+  - docker
+
+before_install:
+  - cd angular-rails-app && docker build -t rails-app .
+  - docker ps
+
```
rdlester: Default install runs bundler locally. We should override it to ensure bundle only runs in the docker container.

angular-rails-app/config/database.yml
```diff
@@ -33,8 +33,10 @@ development:
 # re-generated from your development database when you run "rake".
 # Do not set this db to the same as development or production.
 test:
-  <<: *default
-  database: heatseek_test
+  adapter: sqlite3
+  database: db/test.sqlite3
+  pool: 5
+  timeout: 5000
```
rdlester, line 39: Nice. This should be brought over to my branch.

I assume this automatically handles creating a new database if none already exists?

rdlester (approved):

.travis.yml
rdlester: Or rather, this should be install instead of before_install.

angular-rails-app/Gemfile
chrishenry: 👍

angular-rails-app/config/database.yml
chrishenry: Sadly, it does not. It's just the config.

You still have to go through the `rake db:migrate`, However, I've wrapped that up in the `entrypoint_test.sh` file. So as you add more tests, the command to run them can be added there.

Issue #81, opened 2017-02-13
Figure out how to set up karma with rails sprockets


LKWLaLa:
## Problem ##
We need to write tests for our Angular code.  Currently Karma only identifies Angular dependency injections if the DI module's code is located in our project's main file tree.   However, several dependency injections (such as `templates` and `ui.bootstrap`) were installed via the Gemfile/bundler, and are loaded through the Rails asset pipeline / sprockets.  In order to write tests on the Angular side, we need Karma to be able to load these files as well.  

## Solution ##

Some possible ideas here:   http://stackoverflow.com/questions/18244320/how-to-integrate-karma-with-rails-asset-pipeline

- Currently karma and jasmine config files are located in  `./vendor/assets/bower_components`.  There is also a `karma.conf.js` file in the root directory.  If you think this file structure can be cleaned up/consolidated in a better way, feel free.

rdlester:
https://github.com/SebastianZaha/karma-sprockets
Seems like that should take care of everything? Will attempt to test.
Are all javascript files currently packaged by sprockets?

rdlester:
Nope, karma-sprockets can't resolve javascript from ruby gems.

https://github.com/treycordova/sprockets-loader definitely can, but we'd need to use https://github.com/webpack-contrib/karma-webpack to do so. Hopefully not too big a deal but certainly more config / deps.

rdlester:
The default nodejs provided by apt is deprecated / no longer supported, and webpack requires node v4 or later (due to use of es6 stuff). We have to use a different PPA. Probably should use node 6.x, as it's LTS.

curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash -
sudo apt-get install -y nodejs

rdlester:
Side note: given that webpack sprockets-loader appears to be our only option, there's no reason to use bower.

rdlester:
Another possible (maybe a lot easier) route is to just use a ruby / rails javascript test runner. "Teaspoon" seems to be the runner of choice.

rdlester:
Yeah, Teaspoon seems like the least-complicated option, that and rails-assets lets us avoid needing node altogether.

PR #113, opened 2017-03-05
Remove Rails JS deps from repo; support Rails JS tests via Teaspoon

### LINKED ISSUE #111: Install deps via bower and remove from repo
Pretty much every one of our JS deps is copied into the repo now, which does not seem like the right way to do things. These deps should be installed manually via bower by all users, and should be a part of the Docker web image build.

rdlester:
Rails tests can be run with:
docker-compose run web "export RAILS_ENV=test && bundle exec teaspoon"

Tests go under angular-rails-app/spec/javascripts, must end with _spec.js.

Fixes #111 and #81


rdlester:
Rebasing on the Travis PR. Updated entrypoint_test.sh to run JS tests. Travis should be passing now.

PR #120, opened 2017-03-23
Load Property and Owner data before state change

### LINKED ISSUE #41: Properties page: Have initial state
When loading the properties page, a blank template (with no property information) is shown. Instead of showing this, create an initial loading state that will give the user some feedback that there's still work being done. Once data is loaded, render the property template with data.

LKWLaLa:
Closes #41 

## Changes ##

- Add a `resove` block in `app.js` to fetch data from PropertyService and OwnerService, before the state change and page load.  This way, the user will not have to wait for data after the page renders.  The data will already be there.  This, in combination with the new addition of the spinner on the home page, should solve the issue.

### DIFF
--- a/angular-rails-app/app/assets/javascripts/app.js
+++ b/angular-rails-app/app/assets/javascripts/app.js
@@ -19,17 +19,22 @@ angular
         .state('owner', {
           url: '/owners/:id',
           templateUrl: 'owner/owner.html',
-          controller: 'OwnerController as vm'
+          controller: 'OwnerController as vm',
+          resolve: {
+            owner: function (OwnerService, $stateParams) {
+            return OwnerService.getOwner($stateParams.id);
+          }
+         }
         })
         .state('property', {
           url: '/properties/:id',
           templateUrl: 'property/property.html',
           controller: 'PropertyController as vm',
           resolve: {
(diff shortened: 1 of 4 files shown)


angular-rails-app/app/assets/javascripts/app.js
```diff
@@ -19,12 +19,22 @@ angular
...
         .state('property', {
           url: '/properties/:id',
           templateUrl: 'property/property.html',
-          controller: 'PropertyController as vm'
+          controller: 'PropertyController as vm',
+          resolve: {
+            property: function (PropertyService, $stateParams) {
+            return PropertyService.getProperty($stateParams.id);
+          }
+         }
```
rdlester, line 37: Docs suggest "resolve" is basically dependency injection, so this change should be useful for testing (when you'll want to inject functions that return fake data)!

rdlester (approved):

angular-rails-app/app/assets/javascripts/app.js
LKWLaLa: Interesting point!

LKWLaLa: Thanks for reviewing for me!

Issue #42, opened 2017-01-03
Create paging for objects associated with Properties


chrishenry:
Many properties have thousands of associated objects, which makes returning a single property from the `/properties` endpoint very cumbersome and slow, as well as displaying 100s or 1000s of objects on the page awkward. 

On the front end, implement paging for the hidden details tables.

On the backend, build out more detailed endpoints for hpd_violations, litigations, owners, dob_violations, dob_permits, complaint311s, and hpd_complaints.

To keep the existing frontend snappy, I've implemented limits for most of the `has_many` relationships via the `PropertySerializer`

PR #122, opened 2017-04-03
Pagination of Property details

LKWLaLa:
References Issue #42 

## Changes ##

- Implemented [angularUtils](https://github.com/michaelbromley/angularUtils/tree/master/src/directives/pagination#working-with-asynchronous-data) pagination directive, to paginate details tables (litigations, 311 data, etc) on a Property's show page, on the front end.  

## Notes/Questions ##

While it is possible to use this directive with asynchronous data (rather than loading all data into the front end, and then paginating from there), I am not sure how to paginate **associated objects**  through a has_many relationship  on the back end.   For now, I have simply increased the limit on the has_many relationships in the `PropertySerializer`.   This 50% solves issue #42, but I'm still not certain how to handle the back end.  

I'm wondering if a solution would be to overhaul / restructure the serializers altogether so there is no nesting, but each association has it's own endpoint....  Then to render a show page for a Property, it would require calls to all of the individual endpoints, filtered by property_id.  An endpoint would then get called with a property_id parameter and a page parameter, I guess.    

If anyone has input / thoughts / suggestions, let me know.

### DIFF
--- a/angular-rails-app/app/serializers/property_serializer.rb
+++ b/angular-rails-app/app/serializers/property_serializer.rb
@@ -33,23 +33,23 @@ class PropertySerializer < ActiveModel::Serializer
   end
 
   has_many :litigations do
-    @object.litigations.limit(5)
+    @object.litigations.limit(30)
   end
 
   has_many :dob_violations do
-    @object.dob_violations.limit(5)
+    @object.dob_violations.limit(30)
   end
 
   has_many :dob_permits do
-    @object.dob_permits.limit(5)
+    @object.dob_permits.limit(30)
   end
 
   has_many :complaint311s do
-    @object.complaint311s.limit(5)
+    @object.complaint311s.limit(30)
   end
 
   has_many :hpd_complaints do
-    @object.hpd_complaints.where(status: 'OPEN').limit(5)
+    @object.hpd_complaints.where(status: 'OPEN').limit(30)
   end
 
 end
--- a/angular-rails-app/app/assets/javascripts/app.js
+++ b/angular-rails-app/app/assets/javascripts/app.js
@@ -1,5 +1,6 @@
 angular
-    .module('app', ['ui.router', 'templates', 'google.places', 'angularSlideables', 'ui.bootstrap', 'angularSpinners'])
+    .module('app', ['ui.router', 'templates', 'google.places', 'angularSlideables', 
+      'ui.bootstrap', 'angularSpinners', 'angularUtils.directives.dirPagination'])
     .config(function ($stateProvider, $urlRouterProvider) {
 
       $stateProvider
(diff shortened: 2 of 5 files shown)


angular-rails-app/app/serializers/property_serializer.rb
```diff
@@ -33,23 +33,23 @@ def dob_violations_count
...
 
   has_many :complaint311s do
-    @object.complaint311s.limit(5)
+    @object.complaint311s.limit(30)
   end
 
   has_many :hpd_complaints do
-    @object.hpd_complaints.where(status: 'OPEN').limit(5)
+    @object.hpd_complaints.where(status: 'OPEN').limit(30)
   end
```
rdlester: If I understand what's going on here correctly, this is going to make things worse, as obtaining more of each record type will make requests take longer.

rdlester:
Seems fine on the client side, but we still need to actually paginate things on the server side, which is kind of the actual issue behind #42.

I think the issue is in the Property Serializer.

Take this with a grain of salt, as I know almost zero rails, but I think you'll need to:
- Remove the "has_many" statements from the properties serializer (I think this is what you mean when you talk about nesting).
- Create new routes for each data type.
- Create controllers for those types that look up values by property ID and return a subset according to a page ID (they'll take two params)
- Add corresponding methods to PropertyService.
- Make angular-utils-pagination work with the new PropertyService methods.

I found [an angular package for pagination](https://github.com/begriffs/angular-paginate-anything) that has [a corresponding rails gem](https://github.com/begriffs/clean_pagination), maybe that's worth looking into?

I found [a Gem for Rails pagination](https://github.com/kaminari/kaminari), but it appears to depend heavily on ActiveView, which we've replaced with angular. I did find [a blog post](https://www.tejusparikh.com/2014/pagination-in-angular-js.html) about connecting the two but it seems more complicated than the angular-rails pair above.

chrishenry:
Hey! Thanks for setting up the front end!

After taking a look, I think we should probably move in the direction of an endpoint per model. Then the frontend would be responsible for making a request per relation, filtering by property id.

The tradeoff is more http requests to fully populate the property page, instead of pulling a single json blob. This can work in our favor, cutting down requests, if we're able to only request related objects as the users scrolls down.

Is the directive you added capable of taking a data source / URL as a backend?

LKWLaLa:
@chrishenry @rdlester   Thanks so much, guys!!  Yes, and yes.  Ok, cool.  You've confirmed what I was thinking, but before completely restructuring how the endpoints are set up on the back end, I wanted to get your feedback.  Chris, I wasn't sure about whether it would be a good idea to rely on multiple http requests for the property page, but you're right.... actually that might speed things up for us.  That makes sense.   Regarding the directive I used - it's function is really as a replacement for an `ng-repeat`, so it simply helps with filtering/ rendering the data on the page.  I would still have to write the code to handle the http requests myself, and communicating with the back end.  But I think now I have a pretty good idea of what I need to do.  Thanks so much for your input!  Appreciate it!!  😀

LKWLaLa:
@chrishenry I'm closing this PR, as I've made a new branch.  Will reopen a new PR when finished with the changes we discussed above.

PR #123, opened 2017-04-06
Added hpd-complaints-problems to imports

rdlester:
We missed one last data set from JP's notebooks. This contains the actual complaint details for HPD complaints.


chrishenry:
👍

PR #124, opened 2017-04-06
Add BBL column to dob_permits

rdlester:
dob_permits had a boro name column we cleaned to a 2-letter code, but no numerical ID column. The new code adds this column, populates it from the code using a lookup table, then runs the existing BBL creation code.

### DIFF
--- a/data-imports/clean_utils.py
+++ b/data-imports/clean_utils.py
@@ -51,12 +51,43 @@ def clean_boro(table, column, replacements):
     """
     return '''
     UPDATE {table} SET {column} = preg_replace('/{mn}/i', 'MN', {column});
-    UPDATE {table} SET {column} = preg_replace('/{bk}/i', 'BK', {column});
-    UPDATE {table} SET {column} = preg_replace('/{si}/i', 'SI', {column});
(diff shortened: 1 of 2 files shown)


data-imports/clean_utils.py
```diff
@@ -51,13 +51,44 @@ def clean_boro(table, column, replacements):
...
+    """
+    columnid = column + 'id'
+    sql = ''
+    lookup_table = 'boroid_lookup'
+    if not table_exists(lookup_table):
+        sql = sql + """
+        CREATE TABLE {table} (
+            boro_code varchar(2),
+            boro_id int,
+            PRIMARY KEY(boro_code)
```
chrishenry, line 73: Would it be helpful to have the full borough name here?

rdlester: Not really - we're already running all the boro columns through the "clean_boro" SQL above, which replaces everything with the two-letter codes.

This table is also only meant to be used for this particular step, so it could probably be a temporary table, though it's not really taking much space as is.

chrishenry:
👍

PR #126, opened 2017-04-08
Include housenumber when obtaining addresses from hpd_buildings during r_properties import

### LINKED ISSUE #114: Properties are missing building number in street_address column
In order for our property search to work with the Google API, we need to make sure street addresses have a building number to search by.  Currently, for example, the address `123 2nd Avenue` would appear as `2 Avenue` in the API, and cannot be found.

rdlester:
Fixes #114


chrishenry:
👍

PR #128, opened 2017-04-10
Basic Luigi setup

### LINKED ISSUE #88: data-import Docker container
Right now, I'm running the data-import scripts on Docker by connecting to the nb Docker container (using docker exec -it <container_id> /bin/bash) and manually running the scripts one by one.

We need:
1. A script that runs all the import scripts. This, ideally, should gracefully handle interruptions and automatically pick up from wherever it left off. (This functionality might need to be added to the scripts)
2. A container that, as a part of setup, runs this master script.

rdlester:
Fixes #88.

A very basic Docker image that runs a simple wrapper around our existing import scripts using Luigi.

The "right" way to do Luigi is to start up a scheduler server (luigid), then run tasks against it. Since the current goal of this image is just to run the data imports, I just had the image run the tasks directly without starting up a scheduler server. We might want to actually switch this to a server once we need to do more complex things.

Right now, we don't get too much out of Luigi. It does keep track of which tasks (imports) have run and only runs those it needs to, and can run them in parallel, nicely . However, it'll be very useful when adding incremental imports, or merging data sets together for analysis or exposure in an API (like what the db_connector tasks do now).

One issue with the current setup is that it will only run the imports once. Some scripts (though which I'm not entirely sure) import all available data, so re-running them might pull in new data. You can drop the marker table that Luigi uses to track the imports or specific rows in it to force new updates, but that's a little complicated. Not an issue after incremental update is added.


README.md
```diff
@@ -19,7 +19,8 @@ Before using any of our Docker images, you'll need to create a .env file. The re
 
 * db: The MySQL database storing NYC city data both in semi-raw form and joined into a model for the Rails website. The database can be examined with `docker exec deltanycheatseek_db_1 mysql -u <user from .env> -p <password from .env>`. Local clients can connect at 127.0.0.1:3306 ([localhost will not work](http://stackoverflow.com/a/32361238/103315)).
 * db-udf: Installs [user-defined functions for making regular expression queries](https://github.com/mysqludf/lib_mysqludf_preg) required for data-cleaning into MySQL. Only needs to be run once.
-* nb: Installs Python and starts a Jupyter notebook at [localhost:8888](http://localhost:8888). Can also be used to run the data-import scripts: run `docker exec -it deltanycheatseek_nb_1 /bin/bash`, then execute the scripts
+* luigi: Runs all import scripts using [Luigi](http://luigi.readthedocs.io). Will skip imports already run by the container. Imports can be run in parallel by adjusting the LUIGI_WORKERS parameter in your .env file.
+* nb: Installs Python and starts a Jupyter notebook at [localhost:8888](http://localhost:8888).
```
chrishenry: I think we need one more note here that both the `luigi` and `nb` containers are for running imports, just in a different way. Generally, separate services have separate dependencies, code etc so seeing two services with the same requirements and code threw me a bit.

Next time we meet, should get a sense of who might still be using the nb container for notebooks. If no one's really using it, it might be worthwhile to kill it, and rely on the luigi container.

docker-compose.yml
```diff
@@ -50,3 +50,22 @@ services:
...
+      dockerfile: Dockerfile.luigi
+    env_file: .env
+    # 2G Limit
+    mem_limit: 2147483648
+    environment:
+      MYSQL_HOST: db
+    volumes:
+      - ./data-imports/:/root/data-imports/
+      - ./data/:/root/heatseek/
+      - ./luigi/:/root/luigi/
```
chrishenry: Why this volume? `start.sh` is being copied in.

docker-compose.yml
```diff
@@ -50,3 +50,22 @@ services:
...
+    ports:
+      - "8082:8082"  # default luigid port
```
chrishenry: Hrmm, I'm not getting anything on localhost:8082, and it could be useful.  

Maybe make an issue? Could be a neat improvement.

rdlester: Whoops. Left over from when I was actually starting luigid up and needed to expose a port.

This PR is not really the right way to do things, it's just the first thing I got working. I figured it was better to get something out there / get feedback than get something totally polished.

I didn't go with luigid here b/c I'm still a little confused about the central scheduler and wanted to get something checked in. The model is described here: http://luigi.readthedocs.io/en/stable/execution_model.html
I don't understand if the best approach is to create one container that runs luigid, then create other containers for actually running the jobs (as workers)? Or to just docker-compose exec commands right on luigid?

Thinking about it more, I'm pretty sure it's the former, but then you've got multiple python containers running. However, if I understand the linked doc correctly, the luigid server would only need to install luigi as requirement, and doesn't need the packages required to run specific jobs. Additionally, instead of the "LUIGI_WORKERS" env setting here, you could just use docker-compose scale to spin up multiple worker containers all running the same job. I'll try this out and come back with the results.

README.md
rdlester: Someone at the meetup was running notebooks from nb to get started on stuff.

luigi and nb do have different requirements, though - they use different requirements.txts.

docker-compose.yml
rdlester: Whoops. Left over from when I was running luigid and needed to specify log and state directories.

.dockerignore
```diff
@@ -1 +1,2 @@
 mysql-data/*
+data/*
```
rdlester, line 2: Ah, this would explain why it was taking so long to rebuild docker images, thanks.

README.md
chrishenry: Ah, got it. I see. the requirements.txt at the root of the repo is for the nb container, and the one in data-import is for luigi.

docker-compose.yml
chrishenry: Ya, I agree, just throwing this into a PR is a big step forward.

Looking forward to see what you come up with!

rdlester:
I figured the central scheduler thing out. It works well.

Only goofy thing is that Ctrl-C or `docker stop`-ing the import worker won't actually shut down the worker gracefully (1. it only responds to SIGUSR1, 2. after receiving SIGUSR1 it continues running until its current task is finished). This has two implications:

1. Workers typically hold a lock on a pidfile; if improperly shut down, this lock won't get cleaned up. As such, I'm setting --no-lock when starting a luigi worker.
1. The scheduler still sees whatever task the worker was running as running after the worker is shut down until it times out. Timeout can be configured with [scheduler] -> worker-disconnect-delay; it defaults to 60 seconds. Nothing breaks here, and no tasks will get double-scheduled or go permanently un-scheduled b/c of this, but it does mean the web view will be misleading for a bit after a stop and, to successfully finish all imports, another worker will need to be kept around (even if not running) until the stopped task times out and the scheduler can reschedule it.

rdlester:
I ran all imports successfully using the new setup.

On my computer (high-end laptop, quad-core / 8 threads) with 6 workers, the whole thing took maybe 15-30 mins (wasn't keeping track of exact timing, sorry).

A bit more work is still required to get task history working.

chrishenry:
Awesome. I say we merge this in so it gets into others' hands, and just make an issue for anything you feel next steps would be.

Issue #77, opened 2017-02-04
Refactor: Replace all non-digit chars in clean_bbl


chrishenry:
We're currently replacing a few different characters in `clean_utils.clean_bbl` before consuming `block` and `lot` columns. We should just remove all non-digit characters.

rdlester:
For @virgilib.

rdlester:
The function to modify is at https://github.com/chrishenry/deltanyc-heatseek/blob/master/data-imports/clean_utils.py#L84

virgilib:
@rdlester, in my commit on the pull request, I modified clean_bbl instead of the L84 you mentioned.  Let me know if I misunderstood the issue.

chrishenry:
Hi @virgilib! You were definitely in the right place. (The line number shifted since the comment was made.)

PR #129, opened 2017-04-11
in clean_bbl, replaced all non-digit chars with empty string

virgilib:
fixes issues #77

### DIFF
--- a/data-imports/clean_utils.py
+++ b/data-imports/clean_utils.py
@@ -138,13 +138,12 @@ def clean_bbl(table, boro, block, lot):
 
     return sql + '''
     UPDATE {table} SET {block} = REPLACE({block}, ' ', 0);
-    UPDATE {table} SET {block} = REPLACE({block}, '`', '');
     UPDATE {table} SET {block} = preg_replace('/[^0-9]/', '', {block});
     UPDATE {table} SET {lot} = REPLACE({lot}, ' ', 0);
     UPDATE {table} SET {lot} = preg_replace('/[^0-9]/', '', {lot});
+    UPDATE {table} SET {boro} = preg_replace('/[^0-9]/', '', {boro});
     UPDATE {table} SET bbl =
-            concat(trim({boro}), trim(LPAD({block}, 5, '0')), trim(LPAD({lot}, 4, '0')))
-            WHERE {boro} REGEXP '^[0-9]+$' AND {block} REGEXP '^[0-9]+$' AND {lot} REGEXP '^[0-9]+$';
+            concat(trim({boro}), trim(LPAD({block}, 5, '0')), trim(LPAD({lot}, 4, '0')));
     ALTER TABLE {table} ADD INDEX (bbl);
     '''.format(table=table, boro=boro, block=block, lot=lot)
 
--- a/data-imports/clean_utils.py
+++ b/data-imports/clean_utils.py
@@ -139,7 +139,9 @@ def clean_bbl(table, boro, block, lot):
     return sql + '''
     UPDATE {table} SET {block} = REPLACE({block}, ' ', 0);
     UPDATE {table} SET {block} = REPLACE({block}, '`', '');
+    UPDATE {table} SET {block} = preg_replace('/[^0-9]/', '', {block});
     UPDATE {table} SET {lot} = REPLACE({lot}, ' ', 0);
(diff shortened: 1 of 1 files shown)


data-imports/clean_utils.py
```diff
@@ -139,7 +139,9 @@ def clean_bbl(table, boro, block, lot):
     return sql + '''
     UPDATE {table} SET {block} = REPLACE({block}, ' ', 0);
     UPDATE {table} SET {block} = REPLACE({block}, '`', '');
+    UPDATE {table} SET {block} = preg_replace('/[^0-9]/', '', {block});
```
chrishenry: Great! However, I'd remove the 2 lines above, since the regex takes care of them both!

data-imports/clean_utils.py
```diff
@@ -139,7 +139,9 @@ def clean_bbl(table, boro, block, lot):
...
     UPDATE {table} SET {lot} = REPLACE({lot}, ' ', 0);
+    UPDATE {table} SET {lot} = preg_replace('/[^0-9]/', '', {lot});
```
chrishenry: Same here. The replace of `' '` is now covered by the regex.

chrishenry (changes requested):
Thanks for submitting! Just remove the now superfluous replacements above yours and we'll get this merged!

data-imports/clean_utils.py
```diff
@@ -139,7 +139,9 @@ def clean_bbl(table, boro, block, lot):
...
     UPDATE {table} SET bbl =
             concat(trim({boro}), trim(LPAD({block}, 5, '0')), trim(LPAD({lot}, 4, '0')))
             WHERE {boro} REGEXP '^[0-9]+$' AND {block} REGEXP '^[0-9]+$' AND {lot} REGEXP '^[0-9]+$';
```
rdlester: Also - I'm pretty sure the REGEXP statements down here are now superfluous.

chrishenry: Yep, the whole `WHERE` statement looks like it can go.

data-imports/clean_utils.py
virgilib: Sorry if I misunderstood, but doesn't the line above replace ' ' with zero, therefore not covered by the regex?

data-imports/clean_utils.py
virgilib: To get rid of the first part of the WHERE statement, should I add a preg_replace for the boro column?

rdlester: Yep.

data-imports/clean_utils.py
rdlester: That's a good point, it's not equivalent to your new cleaning. However, it's doing something pretty odd - replacing all spaces with 0s. I would assume this would only be correct if numbers are consistently entered with spaces instead of 0s; if spaces are accidents, then we want to actually remove them entirely. The latter is what I'd expect, but we should investigate if the former ever applies.
I'd prefer we figure this out now; however, if you don't have much time and would rather finish this PR up, we can file an issue for it and deal with it later.

virgilib: I don't have a whole lot of time but seems this issue isn't extremely time-sensitive either, so we have some leeway.  Let me look into the databases this weekend to see how easy it is to resolve.

chrishenry: Yea, it actually is super unclear as to whether the `'  '` -> `0` replacement is valid. 

My suspicion is the spaces are the result of bad scans or manual errors. A good place to start would be to find (from the initial import) some entries with spaces, and then compare the address against a different dataset. Or even use the [NYC BBL tool](http://webapps.nyc.gov:8084/CICS/fin1/find001I)

What would useful is finding if these addresses exhibit a pattern. As in, if all spaces are just that (empty space), then we could simply close them up and the regex is valid. However, if there are cases where spaces represent missing, unknown digits, we may want to mark these as having invalid data. We could then add an additional step to run these through an additional step that overwrites the BBL with data from the NYC BBL tool.

And yep, no particular rush on this...

rdlester:
This was updated. I'll call it good and open a separate issue to investigate the ' ' -> 0 mapping.

rdlester (approved):

virgilib:
close #77

PR #130, opened 2017-04-14
Async paging for objects associated with Properties

### LINKED ISSUE #42: Create paging for objects associated with Properties

LKWLaLa:
closes #42 

## Changes ##

 - This includes front end changes referenced in PR #122 
In addition:
- Implementation of back end pagination, with separate endpoints for each model
- Addition of meta data to all back end paginated endpoints / JSON responses
- Addition of separate API calls to each model associated with a Property.  This is implemented in `property.controller.js` and `property.service.js`
- Tables on front end now fetch data asynchonously
- Small changes to display order of data

## Notes ##

- Back end pagination via will_paginate gem (already installed by Chris)
- Front end pagination via [angularUtils](https://github.com/michaelbromley/angularUtils/tree/master/src/directives/pagination#working-with-asynchronous-data)


angular-rails-app/app/assets/javascripts/property/property.controller.js
```diff
@@ -8,5 +8,98 @@ PropertyController.$inject = ['PropertyService','property'];
...
   var vm = this;  
   vm.data = property.data;
+  vm.itemsPerPage = 10;
+
+  vm.hpdViolations =[];
+  vm.hpdComplaints =[];
+  vm.dobPermits =[];
+  vm.dobViolations =[];
+  vm.litigations =[];
+  vm.complaint311s =[];
```
rdlester: Nit: Seems like it'd be useful to initialize the counts and page numbers as well. Relatedly, you might be able to cut down on some duplication by making a PagedData struct, containing the data (an array), the total count, and the current page number.

angular-rails-app/app/assets/javascripts/property/property.controller.js
```diff
@@ -8,5 +8,98 @@ PropertyController.$inject = ['PropertyService','property'];
...
+
+  vm.hpdViolationsPageChanged = function(newPage) {
+    getHpdViolations(newPage);
+  };
```
rdlester: Nit: Why is there indirection here? Eg. why is there both `vm.hpdViolationsPageChanged` and `getHpdViolations` as opposed to just `vm.getHpdViolations`? It's been a while since I've really done Javascript, so maybe there's some wacky JS thing I'm missing.

If you need to keep both, you can shorten these particular lines to:
`vm.hpdViolationsPageChanged = getHpdViolations;`
That goes for all of these.

angular-rails-app/app/assets/javascripts/property/property.controller.js
```diff
@@ -8,5 +8,98 @@ PropertyController.$inject = ['PropertyService','property'];
 function PropertyController(PropertyService, property) {
   var vm = this;  
   vm.data = property.data;
+  vm.itemsPerPage = 10;
```
rdlester: Super nit: maybe worth renaming vm.data to vm.property to keep things clear? Unless `.data` is something special in Angular.

angular-rails-app/app/assets/javascripts/property/property.controller.js
```diff
@@ -8,5 +8,98 @@ PropertyController.$inject = ['PropertyService','property'];
...
+      vm.complaint311sPage = result.data.meta.current_page
+      });
+    } 
+
+    getComplaint311s(1)
+    getDobPermits(1)
+    getDobViolations(1)
+    getHpdComplaints(1)
+    getHpdViolations(1)
+    getLitigations(1)
```
rdlester: Nit: took me a while to figure out what was going on here (initialization of each dataset to the first page), please add a comment.

angular-rails-app/app/controllers/dob_permits_controller.rb
```diff
@@ -0,0 +1,12 @@
+class DobPermitsController < ApplicationController
+
+  def index 
+    if params[:property_id]
+      @dob_permits = DobPermit.where(
```
rdlester, line 5: Can we simplify things here? The only thing that changes in all of these Controllers is the class name (or at least I think that's a class). I'm not really familiar with Ruby, but I believe you can pass classes around like any other values or objects. I suggest making one controller superclass (PropertyRelatedDataController or something) for all these endpoints, with an unimplemented method for obtaining the Model class to call methods on; subclasses then only need to override that method and return the corresponding class.

I used a similar trick in Python in my current Luigi PR (in luigi_master.py, passing around Python modules).

angular-rails-app/app/controllers/complaint311s_controller.rb
```diff
@@ -0,0 +1,12 @@
+class Complaint311sController < ApplicationController
+
+  def index 
+    if params[:property_id]
+      @complaint311s = Complaint311.where(
+        :property_id => params[:property_id]).paginate(:page => params[:page])
+    end
+    @complaint311s ||= Complaint311.paginate(:page => params[:page])
+    render json: @complaint311s, adapter: :json, 
+    meta: meta_attributes(@complaint311s), status: 200
```
rdlester: Super nit: if a single statement continues across multiple lines, please indent the following lines at least one indent, preferably two. Took me a while to figure out that `render json: ...` and `meta: meta_attributes(...` were a command (`meta:` being an argument to `render`), indentation would make that immediately obvious.

angular-rails-app/app/assets/javascripts/property/property.service.js
```diff
@@ -9,6 +9,30 @@ angular
...
+    return $http.get(API_URL + '/hpd_complaints?' + 'property_id=' + id + '&page=' + pageNumber)
+  }
+
+  this.getHpdViolations = function (id, pageNumber){
+    return $http.get(API_URL + '/hpd_violations?' + 'property_id=' + id + '&page=' + pageNumber)
+  }
+
+  this.getLitigations = function (id, pageNumber){
+    return $http.get(API_URL + '/litigations?' + 'property_id=' + id + '&page=' + pageNumber)
+  }
```
rdlester: Super nit: you could condense this a little by parameterizing the endpoint name (eg. `function(endpoint, id, pageNumber)`), then having the user-facing functions simply call that method (so getHpdComplaints would become just `return getRelatedData('hpd_complaints', id, pageNumber)`.

angular-rails-app/app/assets/javascripts/property/property.html
```diff
@@ -131,6 +134,15 @@ <h4 class="table-title">Open HPD Violations</h4>
...
             </table>
+            <div class="text-center">
+                <dir-pagination-controls
+                pagination-id="hpd_violations"
+                on-page-change="vm.hpdViolationsPageChanged(newPageNumber)"
+                max-size="10"
+                direction-links="true"
+                boundary-links="true" >
+              </dir-pagination-controls>
+            </div>
```
rdlester, line 145: There's a lot of repetition here: let's make a "paged property related data" component and reuse it. I think you can inject the specific bindings (td elements) via "transclusion". The other parameters can just be provided via the isolate scope.

rdlester (changes requested):
Hey, looks great! The logic is definitely correct. My comments are all pretty much on style - there's a bunch of places where you could extract helpers or components, which would really simplify things and cut out a lot of repetition. I think this is a good thing to do and would simplify any future changes, but don't feel obligated if you don't think you have time. And, of course, feel free to disagree.

angular-rails-app/app/assets/javascripts/property/property.controller.js
```diff
@@ -8,5 +8,98 @@ PropertyController.$inject = ['PropertyService','property'];
...
+    }
+
+    function getComplaint311s(pageNumber) {
+    return PropertyService.getComplaint311s(vm.data.id, pageNumber)    
+    .then(function(result) {
+      vm.complaint311s = result.data.complaint311s;
+      vm.totalComplaint311s = result.data.meta.total_count
+      vm.complaint311sPage = result.data.meta.current_page
+      });
+    }
```
chrishenry: There seems to be a ton of code duplication here. Is there any reason you couldn't pass in the type of data you're working with, and make a single generic function? Asked another way, is there any difference in the way 311s, Litigations, etc behave?

angular-rails-app/app/assets/javascripts/property/property.service.js
chrishenry: Yep, I agree with this, and it's a bit more than a nit. The more times the code is duplicated, the more likely there is to be a copy-paste error. 

The sense I'm getting is that all these related data types should more or less work the same way. If we want to improve of change functionality across all these types later, we have to do it in {# types} places. Unless there's a strong intention that I don't know about for these types to act differently, I think we should combine the code.

chrishenry (changes requested):
Overall, looks pretty good. I do think we should refactor to fix the code duplication, tho.

Did we already have tests for these behaviours? I don't see any new or changed tests.

angular-rails-app/app/controllers/dob_permits_controller.rb
LKWLaLa: Will definitely keep this in mind!  I think I'm going to leave this particular one as is for the moment, as I anticipate these action blocks will actually be different.  I would like to add some `group by` and `order by` to the queries, which will be different for each Model (or may not get added at all for some Models). However, since many of the tables are missing data, I waited on that part.

angular-rails-app/app/assets/javascripts/property/property.html
LKWLaLa: I'm going to extract this one into another issue, but I will work on it!

LKWLaLa:
@rdlester @chrishenry  Thanks so much for the great feedback!  I've tried to DRY things up a bit, if you'd like to take another look, and I think I addressed most of the comments.  I have not added tests for this (I know we need them), but that is my next major task, I promise.  Essentially I need to write tests for the whole app.  Obviously the late timing of this is not ideal, but I'm new at testing, so it is a learning process for me.  :)  I will also try to extract some of the html into components, as you suggested, Ryan, but I think I'm going to make that into a separate issue for myself.

In light of the above, ready for review again.
Thanks!!

angular-rails-app/app/assets/javascripts/property/property.controller.js
```diff
@@ -7,6 +7,41 @@ PropertyController.$inject = ['PropertyService','property'];
...
+    'hpdViolations' : new Table("hpd_violations"),
+    'litigations' : new Table("litigations")
+  }
+
+  function Table(urlResource) {
+    this.data = [];
+    this.urlResource = urlResource;
+    this.total = undefined;
+    this.page = 1;
+  }
```
rdlester, line 27: Nit: `Table` can be at the top level, instead of buried inside the PropertyController constructor.

angular-rails-app/app/assets/javascripts/property/property.controller.js
```diff
@@ -7,6 +7,41 @@ PropertyController.$inject = ['PropertyService','property'];
...
+
+  vm.getTableInfo = function(pageNumber, tableName) {
+    return PropertyService.getTableInfo(vm.property.id, pageNumber, vm.tables[tableName]['urlResource'])    
+    .then(function(result) {
+      var underscoreName = vm.tables[tableName]['urlResource']; //need for Rails endpoints
+      vm.tables[tableName]['data'] = result.data[underscoreName]; 
+      vm.tables[tableName]['total'] = result.data.meta.total_count
+      vm.tables[tableName]['page'] = result.data.meta.current_page
+    });
```
rdlester, line 36: Define this on the PropertyController prototype, outside of the constructor.

angular-rails-app/app/assets/javascripts/property/property.controller.js
```diff
@@ -7,6 +7,41 @@ PropertyController.$inject = ['PropertyService','property'];
...
+  }
+
+  function initializeTables(){
+    Object.keys(vm.tables).forEach(function (tableName) {
+      vm.getTableInfo(1, tableName)
+    });
+  }
+
+  initializeTables()
```
rdlester, line 45: `initializeTables` doesn't really need to be its own function; just run the `foreach` directly and comment on the prior line. Readability should also be improved by moving Table and getTableInfo defs out of the constructor.

angular-rails-app/app/assets/javascripts/property/property.html
rdlester: Please do, it would really clean the templates up.

angular-rails-app/app/controllers/dob_permits_controller.rb
rdlester: Well, in theory, even with `group / order by` commands, the structure of the code would still remain the same. You'd just need to inject the particular symbols corresponding to the columns you wish to group or order by along with the model class (and maybe skip group or order if no symbols are provided).

rdlester (commented):
Definitely an improvement, thanks. I still have a few minor comments, though.

angular-rails-app/app/assets/javascripts/property/property.html
chrishenry: Ya, definitely agree with the refactor.

One other thing you could do to make the templates cleaner would be to run the HTML through a beautifier. [A quick google turns up a bunch of different tools.](https://www.google.com/webhp?sourceid=chrome-instant&rlz=1C5CHFA_enUS503US504&ion=1&espv=2&ie=UTF-8#q=html+beautifier)

angular-rails-app/app/assets/javascripts/property/property.controller.js
chrishenry, line 36: Nice, I think this is much cleaner.

chrishenry (approved):
Nice. If you haven't, please make an issue to add tests for this.

LKWLaLa:
@chrishenry @rdlester  Thanks!  I should have time to get back to this again over the weekend.

angular-rails-app/app/assets/javascripts/property/property.html
LKWLaLa: Thanks for the suggestions!  I've made a new issue for the refactor.

angular-rails-app/app/assets/javascripts/property/property.controller.js
LKWLaLa: Thanks!  Yes, there is already an issue regarding testing.  That is the next task on my list, as it is long overdue.

LKWLaLa:
@rdlester I'm not sure I understand the javascript / Angular changes you are suggesting.  Can you show me what you mean?

rdlester (approved):

rdlester:
I guess let's merge this in and I'll open up a follow-up PR.

PR #132, opened 2017-04-16
fix import script permissions

rdlester:
Had to update my git config to allow git to let me change file permissions stored in the repo.


chrishenry:
👍

PR #133, opened 2017-04-17
Infrastructure

chrishenry:
# Changelog

### New

* Initial commit of infrastructure. [Chris Henry]

* Initial commit of infra. [Chris Henry]

### Changes

* Remove mysql port from docker-compose. [Chris Henry]

* Run the web app on port 80 instead of 3000. [Chris Henry]

* Remove hardcoded domain and port from JS files. [Chris Henry]

* Rename google api key for more clarity. [Chris Henry]

### Fix

* Use correct docker file for luigi workers. [Chris Henry]


README.md
```diff
@@ -17,12 +17,12 @@ Before using any of our Docker images, you'll need to create a .env file. The re
 
 `docker-compose up` will build and install our images, and launch them in containers.  These images are:
 
-* db: The MySQL database storing NYC city data both in semi-raw form and joined into a model for the Rails website. The database can be examined with `docker exec deltanycheatseek_db_1 mysql -u <user from .env> -p <password from .env>`. Local clients can connect at 127.0.0.1:3306 ([localhost will not work](http://stackoverflow.com/a/32361238/103315)).
+* db: The MySQL database storing NYC city data both in semi-raw form and joined into a model for the Rails website. The database can be examined with `docker exec deltanycheatseek_db_1 mysql -u <user from .env> -p <password from .env>`.
```
chrishenry, line 20: Note: I removed the mysql port so it wouldn't be exposed when running in production. It does remove a potential way to access the DB locally, but I don't think too many people were connecting that way.

README.md
```diff
@@ -17,12 +17,12 @@ Before using any of our Docker images, you'll need to create a .env file. The re
...
 * luigi-import-worker: Runs all import scripts. Will skip imports already run using this image. Imports can be run in parallel using standard docker methods like `docker-compose scale` (eg. `docker-compose scale luigi-import-worker=4`).
 * nb: Installs Python and starts a Jupyter notebook at [localhost:8888](http://localhost:8888).
-* web: Installs Ruby on Rails and starts a server at [localhost:3000](http://localhost:3000). Like the data-import scripts, you can conect to a running instance with `docker exec -it deltanycheattseek_web_1 /bin/bash` to run rake tasks. To view a demo of the tool, connect to the instance, then run `rake db:setup`. In the demo, fake data will be provided for "10 West 109th Street".
+* web: Installs Ruby on Rails and starts a server at [localhost](http://localhost). Like the data-import scripts, you can conect to a running instance with `docker exec -it deltanycheattseek_web_1 /bin/bash` to run rake tasks. To view a demo of the tool, connect to the instance, then run `rake db:setup`. In the demo, fake data will be provided for "10 West 109th Street".
```
chrishenry, line 25: NOTE: I changed the web port to 80.

angular-rails-app/app/assets/javascripts/home/home.service.js
```diff
@@ -1,4 +1,4 @@
-var API_URL ='http://localhost:3000'
+var API_URL =''
```
chrishenry, line 1: Note: I was a bit conflicted about removing this, but defaulting it to an empty string ensured we used a relative URL that would work in all environments.

We should, however, ensure this is declared only __once__ in the code. Having it in multiple places is a surefire way to ensure there's a bug introduced.

infra/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Google Maps API Key.
+
+The key in use was generated from Chris Henry's Google API Console, but any key
+with Google Maps permissions can be used.
+
+# Setup steps
+
+We should be using luigid, but I can't quite figure out what it's supposed to be
+doing.
```
rdlester, line 34: luigid is a server that handles scheduling and only scheduling. It doesn't know anything about possible tasks.

When a worker starts, it (not the server) pulls a task, computes the dependencies, and checks which are completed. It then submits all incomplete tasks to the scheduler, which then dictates to the worker which task to run.

So, luigi-import-worker is the image that does the actual import work and is the one that needs to be run; luigid is just a server that can orchestrate multiple workers running simultaneously. One potential gotcha is that docker-compose scale doesn't appear to start dependencies of the scaling service (?). I had to `docker-compose up luigid` prior to `docker-compose scale luigi-import-worker=x`. I'd guess the reason is that the scaling of one worker type shouldn't affect the scaling of the others and hence won't even check if the others have workers running at all.

infra/main.tf
```diff
@@ -0,0 +1,16 @@
+provider "digitalocean" {
+  # You need to set this in your .bashrc
+  # export DIGITALOCEAN_TOKEN="Your API TOKEN"
+  #
+}
+
+resource "digitalocean_droplet" "delta_droplet" {
+  ssh_keys           = [8187378]
```
rdlester: Without knowing anything about tf, this seems like something that maybe shouldn't be here?

infra/main.tf
```diff
@@ -0,0 +1,16 @@
...
+  image              = "ubuntu-16-10-x64"
```
rdlester: Maybe it'd be better to stick with an LTS Ubuntu version (16.04)?

infra/terraform.tfstate
```diff
@@ -0,0 +1,48 @@
...
...pt-transport-https\n  - ca-certificates\n  - curl\n\nruncmd:\n  # Install docker-ce\n  - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -\n  - sudo add-apt-repository \"deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable\"\n  - sudo apt-get update \u0026\u0026 apt-get install -y docker-ce\n  - sudo curl -o /usr/local/bin/docker-compose -L \"https://github.com/docker/compose/releases/download/1.11.2/docker-compose-$(uname -s)-$(uname -m)\"\n  - sudo chmod +x /usr/local/bin/docker-compose\n\n  # Install the DO monitoring agent\n  - curl -sSL https://agent.digitalocean.com/install.sh | sh\n\n\n  # Get the app up and running\n  - [mkdir, -p, /opt/apps/deltanyc-heatseek]\n  - git clone https://github.com/chrishenry/deltanyc-heatseek /opt/apps/deltanyc-heatseek\n  - cp /opt/apps/deltanyc-heatseek/.env.example /opt/apps/deltanyc-heatseek/.env\n  - cd /opt/apps/deltanyc-heatseek \u0026\u0026 git checkout infra\n  - cd /opt/apps/deltanyc-heatseek \u0026\u0026 docker-compose up -d\n  - cd /opt/apps/deltanyc-heatseek \u0026\u0026 docker-compose exec nb \"for f in `ls | grep import | grep .py | grep -v pyc`; do python $f; done\"\n\n",
```
rdlester, line 35: Is this setting a bash script to run on the droplet via user_data? This seems kinda wacky. Is it at least possible to add \ns?

chrishenry: Ah, so this isn't the actual file, but a file terraform uses to [manage state](https://www.terraform.io/docs/state/). The actual readable, editable user-data file is at https://github.com/chrishenry/deltanyc-heatseek/pull/133/files#diff-3a719607b63443e4bc05ab2b17c50ee2

infra/main.tf
chrishenry: This _should_ be ok. Briefly, it's the ID of the public key I generated, stored in DO, and will be provisioned in the default account of the image.

However, it's pretty unclear, so I'll def add a comment.

infra/main.tf
chrishenry: I don't see too much benefit there. I'd rather be on a newer version of the OS (which will hopefully ship with newer versions of Docker, etc). Since part of our deploy strategy will be to wholesale replace the server, there shouldn't be too much of a challenge in changing out the image to a newer one.

infra/README.md
chrishenry: Ahh, got it. that's super helpful. So with the workers, do they execute a set number of jobs and then exit? I observed the one worker do a few jobs, and then exit with code 137 (?).

Next round, I'll definitely play around with using luigi to totally automate the process.

rdlester: Not sure why there'd be a 137 exit code - they should exit successfully. My guess is there's an error of some sorts in one of the tasks? Each worker should log the general status of tasks, plus logs from the task (the import script) itself; if these aren't logging to stdout, they can be examined with `docker-compose logs luigi-import-worker` + grepping for the particular worker in question. Luigid also doubles as a web server; accessing it in a web browser shows a page with the status of all tasks. It can save a complete history of executed tasks, but that requires some extra setup I haven't finished yet.

While it may be possible to specify multiple tasks for a worker, our setup runs the worker with a single task. That task, however, is simply a `WrapperTask` that lists all the import scripts as deps. (This is ImportAll in luigi_master) As mentioned, the worker will start up, list out all the imports and check if they've been completed. Luigi then submits the list of tasks to the scheduler, which will send back a task if one exists that is not already in progress. Once the worker finishes (with either success or failure), it tells the scheduler the result and gets a new task (if one exists that is not already in progress).
This means that if you're running multiple workers, the exact number of tasks each worker will run is not set in stone, and will depend on timing. Of course, if you run a single worker, the number of tasks that worker is determined, since that worker has to do all the work.

PR #140, opened 2017-04-18
Db connector fixes

### LINKED ISSUE #131: Many rails tables have empty columns
If you navigate to a property show page in the web app and QA the user interface by clicking to view the tables, you will notice there are many blank columns.  When PR #130 is merged, you will have access to the table endpoints to look at this, but for now the quickest way to see this is to run the app at `localhost:3000` and view a property.  You could try 1749 Grand Concourse, Bronx for starters.

Many tables have blank columns, but for example, here is DOB Permits:
image

rdlester:
Fixes #131 (????)

There are still a few empty columns in r_hpd_complaints; filling these in is a larger task involving joining the complaints and complaints_problems tables. I've opened a separate issue for that.

### DIFF
--- a/angular-rails-app/db/schema.rb
+++ b/angular-rails-app/db/schema.rb
@@ -43,7 +43,7 @@ ActiveRecord::Schema.define(version: 20170225185334) do
     t.datetime "job_start_date"
     t.string   "job_type",        limit: 255
     t.string   "job_num",         limit: 255
-    t.string   "filling_status",  limit: 255
+    t.string   "filing_status",  limit: 255
     t.string   "permit_type",     limit: 255
     t.string   "bldg_type",       limit: 255
   end
--- a/angular-rails-app/lib/tasks/db_connector.rake
+++ b/angular-rails-app/lib/tasks/db_connector.rake
@@ -319,7 +319,7 @@ namespace :db_connector do
 
     puts "Adding indexes..."
     indexes = [
-      ["dob_violations", "violation_type"]
+      ["dob_violations", "violation_type_code"]
     ]
     add_indexes(indexes)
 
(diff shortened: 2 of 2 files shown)


angular-rails-app/db/schema.rb
```diff
@@ -43,7 +43,7 @@
     t.datetime "job_start_date"
     t.string   "job_type",        limit: 255
     t.string   "job_num",         limit: 255
-    t.string   "filling_status",  limit: 255
+    t.string   "filing_status",  limit: 255
```
chrishenry, line 46: So that's why that looked so weird. 😉

angular-rails-app/lib/tasks/db_connector.rake
```diff
@@ -322,25 +324,25 @@ namespace :db_connector do
...
+      "LL6291", # -LOCAL LAW 62/91 - BOILERS",
+      "C", # -CONSTRUCTION",
+      "B", # -BOILER",
     ]
 
     in_list =  "\"" + v_types.join("\", \"") + "\""
 
     puts "Pulling dob violations..."
-    sql = "SELECT * FROM dob_violations WHERE violation_type IN (#{in_list})"
+    sql = "SELECT * FROM dob_violations WHERE violation_type_code IN (#{in_list})"
```
chrishenry, line 345: Nice, this'll probably perform better. You should change the `add_indexes` params above to match this column.

chrishenry (changes requested):

angular-rails-app/lib/tasks/db_connector.rake
rdlester: Done.

rdlester: Done.

chrishenry: sweet. 👍

chrishenry (approved):

PR #142, opened 2017-04-19
Fill hpd_type from registration_contacts.type

rdlester:


rdlester:
Not to be a jerk but I'm not even going to bother with review for something this simple.

rdlester:
This was for #131

PR #144, opened 2017-04-21
Fix Rails DB Schema & Migrations

rdlester:
- Most important: remove useless columns (dob_permits.bin, borough, block, lot - all info is contained in linked Properties table).
- Add bbl to Properties table.
- Condense migrations into single file using [squasher](https://github.com/jalkoby/squasher). Maybe a little controversial, but we don't really need the old migrations as none of them were launched to prod.
- Ensure Migrations match schema.rb (dob_permits.work_type was present in schema.rb but not in the migrations).
- Remove Lindsey's old schema and migrations. `rake db:migrate` seemed to pick them up in addition to Chris' migrations, resulting in a whole bunch of junk getting added to the db and schema.rb.
- Produce a new schema.rb by running the migration.
- Ignore sqlite3 files (ie. the rails test db).
- Limit progress logging / stdout flushing to very slightly optimize db_connector scripts.


LKWLaLa:
@rdlester  The condensed migrations are fine with me, if you have confirmed that this does not break anything.  (It would be useful right now to have more tests...)  Due to the removal and recreation of the whole schema, it is difficult to see which column names you have corrected or removed altogether, though.  Would you mind adding that info to your changelog above for reference?  Thanks so much for tackling this!

rdlester:
@LKWLaLa You can see the removed columns in the diff for schema.rb. Only thing that changed is dob_permits.

I actually caught one erroneously removed column that was added directly to the old schema.rb but not the (new) migrations - dob_permits.work_type. This has been fixed.

I double-checked dob_permits - seems to work fine.

LKWLaLa (approved):
ok!

chrishenry:
👍 

Also, I checked this branch out and ran `db:migrate`, so I'm pretty confident these changes should be fine.

PR #150, opened 2017-04-25
Add bbl to Properties model

rdlester:
Seems like a lot of these db_connector tasks should really be done with sql joins, and directly joining on the bbls seems like the simplest thing. The "BBL" is sort of already present, but not in joined form. Without including the joined form, we'd have to do extra work to bring in tables that have only the BBL and not the individual components in columns, like the rent stabilization table.

Also, limit logging to console to reduce script workload a tiny, tiny bit.

Please note this PR is based off my other current PR. If you merge it, please make sure it's merging to master.

### DIFF
--- a/angular-rails-app/db/migrate/20170225185334_init_schema.rb
+++ b/angular-rails-app/db/migrate/20170225185334_init_schema.rb
@@ -129,10 +129,12 @@ class InitSchema < ActiveRecord::Migration
       t.string   "borough",             limit: 255
       t.integer  "block",               limit: 4
       t.integer  "lot",                 limit: 4
+      t.integer  "bbl",                  limit: 8
       t.integer  "hpd_registration_id", limit: 4
     end
 
     add_index "properties", ["borough", "block", "lot"], name: "index_r_properties_on_borough_and_block_and_lot", unique: true, using: :btree
+    add_index "properties", ["bbl"], name: "index_r_properties_on_bbl", using: :btree
     add_index "properties", ["street_address", "zipcode"], name: "index_r_properties_on_street_address_and_zipcode", using: :btree
 
   end
--- a/angular-rails-app/db/schema.rb
+++ b/angular-rails-app/db/schema.rb
@@ -146,7 +146,7 @@ ActiveRecord::Schema.define(version: 20170225185334) do
   end
 
   add_index "properties", ["borough", "block", "lot"], name: "index_r_properties_on_borough_and_block_and_lot", unique: true, using: :btree
-  add_index "properties", ["bbl"], name: "index_r_properties_on_bbl", using: :btree
+  add_index "properties", ["bbl"], name: "index_r_properties_on_bbl", unique: true, using: :btree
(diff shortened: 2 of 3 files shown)


angular-rails-app/db/migrate/20170225185334_init_schema.rb
```diff
@@ -129,10 +129,12 @@ def up
       t.string   "borough",             limit: 255
       t.integer  "block",               limit: 4
       t.integer  "lot",                 limit: 4
+      t.integer  "bbl",                  limit: 8
       t.integer  "hpd_registration_id", limit: 4
     end
 
     add_index "properties", ["borough", "block", "lot"], name: "index_r_properties_on_borough_and_block_and_lot", unique: true, using: :btree
+    add_index "properties", ["bbl"], name: "index_r_properties_on_bbl", using: :btree
```
chrishenry, line 137: Would it make sense to make this unique?

chrishenry:
Here's what the migration would look like;

```ruby
class AddBblToProperties < ActiveRecord::Migration
  def change
    add_column :properties, :bbl, :integer
    add_index :properties, [:bbl]
  end
end
```

angular-rails-app/db/migrate/20170225185334_init_schema.rb
rdlester: I did make it unique, originally, but it seemed like it janked up the import? However, just tried it again and it seemed to work. Will add.

rdlester:
Would you like for me to put this in a migration instead of modifying the existing tables?

For the reasons mentioned on Slack, I can't really get a good schema.rb after running db:migrate - it's including the schemas of every table in the database in schema.rb, including our raw data tables. That feels like excess noise we don't want to include in schema.rb. (I think the proper fix for this is to move the Rails tables into their own database on the same mysql server?)

angular-rails-app/db/migrate/20170225185334_init_schema.rb
rdlester: Done.

chrishenry:
Yep, don't know if you saw the response in slack, but I totes agree with separating the data and web tables into separate DBs (but same server). I captured the task here. https://github.com/chrishenry/deltanyc-heatseek/issues/152

However, I do think that migrations are going to be more predictable for us in the long term. Both for protecting production data and as the standard way for keeping up schema changes.

rdlester:
Re: migrations: yes, they are better in the long run for sure, I never meant to keep modifying the init_schema migration and schema.rb beyond this sequence of changes / until #152 is fixed.

Do you think it's worth switching over to migrations in these changes now?

chrishenry:
Nah, I think it's fine. I'll take a stab at separating the DBs tomorrow. Then we should be able to move back to migrations and still have schema.rb make sense.

rdlester:
Oh no none of this stuff got merged to master. Will try to fix this up.

rdlester:
Github doesn't really support chained / stacked PRs, which is a huge pain. You have to merge last-to-first, unfortunately.

rdlester:
Never mind, this got merged via the rent-stabilization PR. Sorry this got messed up.

PR #151, opened 2017-04-25
Include rent stabilized counts in Rails properties

rdlester:
Also ensure unit counts are backfilled into properties imported from hpd_buildings, as only pluto includes that data.

Again, note that this PR is based off of the PR adding bbl to r_properties. If merging, please ensure this is merged to master.

### DIFF
--- a/angular-rails-app/db/migrate/20170225185334_init_schema.rb
+++ b/angular-rails-app/db/migrate/20170225185334_init_schema.rb
@@ -129,10 +129,12 @@ class InitSchema < ActiveRecord::Migration
       t.string   "borough",             limit: 255
       t.integer  "block",               limit: 4
       t.integer  "lot",                 limit: 4
+      t.integer  "bbl",                  limit: 8
       t.integer  "hpd_registration_id", limit: 4
     end
 
     add_index "properties", ["borough", "block", "lot"], name: "index_r_properties_on_borough_and_block_and_lot", unique: true, using: :btree
+    add_index "properties", ["bbl"], name: "index_r_properties_on_bbl", using: :btree
     add_index "properties", ["street_address", "zipcode"], name: "index_r_properties_on_street_address_and_zipcode", using: :btree
 
   end
--- a/angular-rails-app/db/migrate/20170225185334_init_schema.rb
+++ b/angular-rails-app/db/migrate/20170225185334_init_schema.rb
@@ -123,13 +123,13 @@ class InitSchema < ActiveRecord::Migration
       t.string   "zipcode",             limit: 255
       t.integer  "total_units",         limit: 4
       t.string   "bin",                 limit: 255
-      t.boolean  "rent_stabilized"
+      t.integer  "rent_stabilized",     limit: 4
       t.datetime "created_at",                      null: false
       t.datetime "updated_at",                      null: false
       t.string   "borough",             limit: 255
       t.integer  "block",               limit: 4
       t.integer  "lot",                 limit: 4
-      t.integer  "bbl",                  limit: 8
+      t.integer  "bbl",                 limit: 8
       t.integer  "hpd_registration_id", limit: 4
     end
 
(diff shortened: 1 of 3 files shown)


angular-rails-app/db/migrate/20170225185334_init_schema.rb
```diff
@@ -123,13 +123,13 @@ def up
       t.string   "zipcode",             limit: 255
       t.integer  "total_units",         limit: 4
       t.string   "bin",                 limit: 255
-      t.boolean  "rent_stabilized"
+      t.integer  "rent_stabilized",     limit: 4
```
chrishenry, line 126: Ohh, have you just been changing the initial migration? And using db:schema:load?

That might explain why you've been seeing some wonkiness when trying to make changes. Lindsey and I have been using migrations to make changes. 

I think we should probably be using migrations, since `db:schema:load` can destroy data in production. [Here's a pretty good rundown](http://stackoverflow.com/questions/5905287/rake-dbschemaload-vs-migrations)

chrishenry (changes requested):

angular-rails-app/db/migrate/20170225185334_init_schema.rb
rdlester: No, this is intentional. There are a couple of problems I'm running into that I posted about in Slack and would like your input on.

- The main blocker is that when I run `rake db:migrate`, schema.rb gets updated with schemas for every table in the database, including the data import tables. This not only results in a bloated schema.rb, it's also wildly confusing and potentially dangerous (since the "r_" prefixes are stripped in schema.rb, there are two "create_table dob_permits" entries, for example). Hence, even if I wrote regular migrations for updating the database, I'd still have to update schema.rb by hand. Because I'm updating it by hand, I'm stuck running `db:schema:load` to make sure I didn't typo.

- Given a) I'm updating schema.rb by hand, b) we don't really have a production site up yet (that's sort of changed since I've started, I suppose), c) these modifications (to r_properties in particular) more or less require I nuke the whole Rails DB anyways (I have to rebuild r_properties, which in turn means I have to rebuild everything else to keep the property_ids correct, unless we add ORDER BY to our INSERT commands), d) we can't really lose data given that everything is pulled from the data-imports and is hence is always reproducible and e) I'm making these changes in very quick succession, it seemed simplest to just update the initial migration in line with the changes made to schema.rb.

Please let me know if:
- There is some Rails config I can set to prevent db:migrate from pulling in all the data-import tables.
- If it's okay to separate the Rails DB into its own DB, thus bypassing the problem.
- whether I should just go straight into migrations regardless of these other issues.

Anyways, the historical context:
In PR #140 , I did things entirely wrong: I just added lines directly to schema.rb and didn't change migrations at all. Of course, I didn't test this at all - I didn't know what the different db rake tasks did, and had just made the database changes manually via SQL commands. I was in a rush to get that PR in, but afterwards that change bugged me, so I read up on the actual system, at which point I ran into the schema.rb bug described above (where data-import tables get pulled in). My first attempt to fix it led to accidentally dropping the entire DB (another reason I think the Rails DB and the data imports DB should be separated). My next attempt was to run squasher and delete Lindsey's old DB and migrations, which was the first PR in this series. This turned up a column present in Lindsey's old migrations but not our current ones, but didn't fix my schema.rb issues.

rdlester: From our conversation in the other thread, I think this is fine for now?

rdlester:
I've pointed the base at master, so we can go ahead and merge. It would be nice to get my series of changes in.

angular-rails-app/db/migrate/20170225185334_init_schema.rb
chrishenry: Yep, let's get this merged. I think we're on the same page about the long term plan is, and what to do in the interim.

If you're curious, there's an [ignore_tables](https://apidock.com/rails/v3.1.0/ActiveRecord/SchemaDumper/ignore_tables/class) option in ActiveRecord. I played with it, to see if would address your issues, but there were weird interactions with the `table_prefix`.

chrishenry (approved):

rdlester:
Fixed #131.

Issue #149, opened 2017-04-23
Optimize db_connector tasks


rdlester:
The db_connector tasks are incredibly slow at the moment, likely due to processing the entire data set in Ruby. Many seem like they could be performed via SQL joins, greatly improving performance.

rdlester:
Not an MVP issue but it's much easier to get stuff done when connector tasks take on the order of seconds instead of hours.

PR #153, opened 2017-04-26
Optimize owners connector task

rdlester:
Goes from taking forever (I didn't actually let it finish) to a few seconds. Also definitely works now, no doubt. 

There still are duplicate entries for the same corporations, though. No idea why HPD is giving them so many registration ids?

First step towards #149


chrishenry:
Ya, this is going to be much faster. 🚤 🐎 

The downside here is we're a bit more prone to error, since we lose the structure & validation provided by the Rails models. However, I think the long term direction is to push as much of that into the data science side.

Issue #101, opened 2017-02-22
Write tests for Angular front end


LKWLaLa:
We currently are not testing our Angular features, and should be! 

In order to begin writing tests however, issue #81 needs to be addressed - this is presently a roadblock.

rdlester:
#81 is fixed, we should do this.

PR #154, opened 2017-04-26
refactored so that API_URL is defined in one place in app.js

### LINKED ISSUE #139: Refactor API_URL in JS
Currently, the API_URL is in 3 separate places in the JS.

This should be refactored so it can be in only 1 place.

virgilib:
fixes #139

### DIFF
--- a/angular-rails-app/app/assets/javascripts/app.js
+++ b/angular-rails-app/app/assets/javascripts/app.js
@@ -1,7 +1,13 @@
+var env = {}
+
+if(window && window.ENV){
+    Object.assign(env, window.ENV);
+}
+
 angular
     .module('app', ['ui.router', 'templates', 'google.places', 'angularSlideables', 
       'ui.bootstrap', 'angularSpinners', 'angularUtils.directives.dirPagination'])
-    .constant('API_URL', '')
+    .constant('ENV', env)
     .config(function ($stateProvider, $urlRouterProvider) {
 
       $stateProvider
--- a/angular-rails-app/app/assets/javascripts/app.js
+++ b/angular-rails-app/app/assets/javascripts/app.js
@@ -1,6 +1,7 @@
 angular
     .module('app', ['ui.router', 'templates', 'google.places', 'angularSlideables', 
       'ui.bootstrap', 'angularSpinners', 'angularUtils.directives.dirPagination'])
+    .constant('API_URL', '')
     .config(function ($stateProvider, $urlRouterProvider) {
 
       $stateProvider
--- a/angular-rails-app/app/assets/javascripts/env.js
+++ b/angular-rails-app/app/assets/javascripts/env.js
@@ -1,0 +1,4 @@
+(function (window) {
+  window.ENV = window.ENV || {};
+  window.ENV.API_URL = '';
+}(this));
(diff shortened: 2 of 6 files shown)


chrishenry:
Sweet refactor! 

Quick question; does this change make `API_URL` into a global? Or asked another way, could we do the same thing, but put the definition of `API_URL` in a script tag in the head? Would that still work with Angular?

Reason I ask is that we want to move to a place where that value can be defined without changing code. As in, someone deploying this could change an environment variable, which would be output in the head, and then picked up by the Angular app.

virgilib:
Ah ok I see what you're saying!  As far as I know, putting the environment variable in the script tag in the head works for server-side variables but wouldn't get it into the client-side JS.  Let me think about it some more.

virgilib:
How about something like this?

chrishenry:
Yea, i think that moves us forward a bit. So, if I were to output a script tag onto the head on every page like;

```html
<script>
ENV = {
    API_URL: "/new/api/url"
}
</script>
```
would that value be passed through?

virgilib:
I don't think so since the env.js approach moves API_URL to an environment file like how variables are defined in .env to allow for the variable to be changed without touching the AngularJS app.

I'll have to think a bit more about how to read the variable from a script tag! Thanks for clarifying.

chrishenry:
@virgilib I think this is a big step forward from what we have, so let's get it merged! Thanks!

Keep thinking about how you'd pass vars through script tags, and let me know if you need any help!

rdlester:
Re @virgilib - pretty sure a script tag with inline JS is exactly equivalent to a script tag with the same JS but in a file. Either way, context is shared across tags, so as long as the variable is set globally (or on the window) other stuff should be able to pick it up.

However, because of this equivalence, I'm not sure why you'd prefer one approach over the other. I think (though it's not stated) that the reason Chris is interested in this is for setting up local / test / prod server endpoints that can be automatically set up according to a flag. I think either approach (env file v. raw script tag) would involve adding a step to compilation, I guess the question is which step is easier? I prefer the JS file, just for the sake of keeping all JS stuff in JS files, but I'm not familiar with Rails so I don't know what the typical setup is.

rdlester:
Also - Sorry, but this is definitely not injecting API_URL into the Services as exists now. Putting "console.log(ENV.API_URL + "/query?")" into the HomeService yields "undefined/query?" which results in all service calls to mysteriously no-op (on the Home page at least, neither the "then" nor the "error" are getting called, only "finally").

I'm going to revert this. We should be better about testing stuff.

chrishenry:
Hmm, shouldn't have tests caught that?

rdlester:
I don't think any real tests are set up for the JS code yet? If so, they probably aren't set up for the Services.

rdlester:
There is one test but it's not doing anything:
https://github.com/chrishenry/deltanyc-heatseek/blob/master/angular-rails-app/spec/javascripts/controllerTest_spec.js

rdlester:
See #101

rdlester:
Sorry for the quick / sudden revert, just want to fully verify everything is working prior prior to showing the MVP.

virgilib:
Ah sorry about this!  My development environment must be set up wrong so my code wasn't getting tested....  Will debug and fix.

Issue #148, opened 2017-04-22
DOB_Permits and HPD_Registrations tasks broken


rdlester:
Both due to mis-entered numerical columns, it seems.

Will add the actual errors soon.

rdlester:
DOB_PERMITS error:
DatabaseError: 1366 (HY000): Incorrect integer value: '' for column 'boroughid' at row 632161
(pretty sure this refers to either the clean_utils for creating boroughids from borough letter codes or the clean_utils for using those ids in creating a BBL column)

rdlester:
HPD_REGISTRATIONS error:
DatabaseError: 1265 (01000): Data truncated for column 'lot' at row 2772
(???)

rdlester:
@nickefimov I've fixed the dob_permits issue since I pretty much knew what was up. I have no idea what's going on with that registrations error though, would much appreciate it if you took a look!

nik-efimov-yyz:
@rdlester how do I run an import script? I haven't worked much with Docker in the past.

nik-efimov-yyz:
@rdlester is this the `db_connector:owners` task? I ran it and it completed successfully..

rdlester:
No; this is `import_hpd_registrations.py` inside of the data-imports folder. `db_connector:owners` is likely a complete no-op unless you've already run the python import to obtain the actual raw data. 

The suggested way to run the the data-imports is to just run the `luigi-import-worker` service via docker-compose. Just `docker-compose up luigi-import-worker`*. However, this service (when run via `up`) runs all the imports and keeps track of which ones it has successfully run, so while it's great for pulling down all the data and getting set up, it's maybe not the best when debugging or adding new functionality to an import. You can run individual scripts by opening a bash shell in a "container" on top of `luigi-import-worker` (`docker-compose run luigi-import-worker /bin/bash`), then just execute the corresponding import python file as a script.

*(alternatively, you can use `docker-compose scale` if you'd like to run the imports in parallel, but you have to run `docker-compose up luigid` manually prior to doing so to start the dependency service that `up` will start automatically)

PR #155, opened 2017-04-26
Fix and optimize dob_permits pipeline

rdlester:
Fixes the dob_permits part of #148 , another notch in #149 (db_connector now takes ~15 secs instead of tens of minutes).


chrishenry:
👍

PR #156, opened 2017-04-27
Fix HPD Registrations import

### LINKED ISSUE #148: DOB_Permits and HPD_Registrations tasks broken

nik-efimov-yyz:
Fixes #148 

The issue was that sometimes LOT data was missing from the source CSV.


rdlester:
Looks good to me.

PR #157, opened 2017-04-28
Fixes #137

### LINKED ISSUE #137: On Ubuntu, concat'ing Pluto files fails with Killed

nik-efimov-yyz:
Fixes #137


chrishenry:
Oh, got it. We pass `dtype` pretty much everywhere else, and that probably cuts down on the amount of work pandas has to do.

Thanks!

PR #163, opened 2017-04-30
Optimize HPD_Complains Rails import

rdlester:
311 is a whole 'nother can of worms, so this can close out #149 .

### DIFF
--- a/angular-rails-app/lib/tasks/db_connector.rake
+++ b/angular-rails-app/lib/tasks/db_connector.rake
@@ -150,6 +150,7 @@ namespace :db_connector do
 
   desc "Pull in hpd complaints"
   task hpd_complaints: :environment do
+    # TODO: missing columns: complaint_type, major_category_id, minor_category_id, code_id
     conn = ActiveRecord::Base.connection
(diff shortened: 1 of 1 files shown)


angular-rails-app/lib/tasks/db_connector.rake
```diff
@@ -150,58 +150,19 @@ namespace :db_connector do
 
   desc "Pull in hpd complaints"
   task hpd_complaints: :environment do
-
+    # TODO: missing columns: complaint_type, major_category_id, minor_category_id, code_id
```
chrishenry, line 153: Ok, I think I finally tracked this down.

These columns aren't a part of `hpd_complaints`, but rather part of `hpd_complaint_problems`. Since there can be more than one problem to a complaint, we need to do some extra work to show this the right way;

* https://github.com/chrishenry/deltanyc-heatseek/issues/165
* https://github.com/chrishenry/deltanyc-heatseek/issues/166
* The design of this section needs to reflect this...somehow.

chrishenry (approved):

PR #161, opened 2017-04-30
Optimize functionality-complete db_connectors

rdlester:
Also add "violation_type_code" to r_dob_violations ("violation_type" seems to now store something much different in the latest data dump).

Two imports remain, but they have other required work: hpd_complaints needs an overhaul to pull in problems; 311 needs to geocode against a local server / library to work around the publicly-hosted geoclient API restrictions.

Also fixes #127

PR #167, opened 2017-05-01
Fix conflicts b/w PRs #158 and #161 / #163

rdlester:
PRs #158 and #161 / #163 got merged without fully fixing everything up beforehand, so some of the db_connectors aren't updated to use the separate databases. This should fix it.


chrishenry:
👍 Thanks!

PR #169, opened 2017-05-02
Merge problems with hpd complaints in Rails

### LINKED ISSUE #166: Import hpd complaint problems into Rails
Currently, we only import hpd complaints into the Rails app.

To show more information about the HPD complaints, we'll need to create a new table to store these, since there are many problems to a complaint.

rdlester:
Obviously doesn't totally work yet, since the problems columns are so messed up, but the query will do what it is supposed to do (I tested with fake problems). Once we have the problems down correctly, we will maybe need to change the Rails schema around to different info or change the column types.

Also checks in the migration-generated schema.

Fixes #166

### DIFF
--- a/angular-rails-app/lib/tasks/db_connector.rake
+++ b/angular-rails-app/lib/tasks/db_connector.rake
@@ -150,14 +150,22 @@ namespace :db_connector do
 
   desc "Pull in hpd complaints"
   task hpd_complaints: :environment do
-    # TODO: missing columns: complaint_type, major_category_id, minor_category_id, code_id
+    indexes = [
+      ["#{ENV['MYSQL_DATABASE_DATA']}.hpd_complaints", "complaintid"],
+      ["#{ENV['MYSQL_DATABASE_DATA']}.hpd_complaints_problems", "complaintid"],
+    ]
+    add_indexes(indexes)
     conn = ActiveRecord::Base.connection
     sql = "INSERT IGNORE INTO hpd_complaints
-        (property_id, received_date, complaint_id, apartment, status, status_date, status_id)
+        (property_id, received_date, complaint_id, apartment, status, status_date, status_id,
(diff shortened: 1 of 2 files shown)


angular-rails-app/lib/tasks/db_connector.rake
```diff
@@ -150,14 +150,22 @@ namespace :db_connector do
...
+        complaint_type, major_category_id, minor_category_id, code_id)
         SELECT p.id, c.receiveddate, c.complaintid, c.apartment, c.status,
-        c.statusdate, c.statusid
+        c.statusdate, c.statusid, prob.type, prob.majorcategoryid, prob.minorcategoryid,
+        prob.codeid
         FROM properties AS p INNER JOIN #{ENV['MYSQL_DATABASE_DATA']}.hpd_complaints AS c
-            ON p.bbl = c.bbl;"
+            ON p.bbl = c.bbl
+        INNER JOIN #{ENV['MYSQL_DATABASE_DATA']}.hpd_complaints_problems AS prob
+            ON c.complaintid = prob.complaintid;"
```
chrishenry, line 168: Hrmm, I'm not sure this is the correct schema. Even though data isn't correct on the `DATA` side, there are more rows in the `problems` table than the `complaints` table, indicating there's one-to-many relationship. Doing it this way will probably cause duplicate complaints.

chrishenry (changes requested):

angular-rails-app/lib/tasks/db_connector.rake
chrishenry: OK, based on [conversation in slack](https://housingdataconsortium.slack.com/archives/G4HS4L118/p1493778990564694), this does seem to be the desired way to present this data.

chrishenry (approved):

PR #172, opened 2017-05-03
HPD Complaints connector update

nik-efimov-yyz:


chrishenry:
Looks OK. I'd argue that we should comment that HTML instead of removing, as we're going to want to eventually implement those.

angular-rails-app/lib/tasks/db_connector.rake
```diff
@@ -125,20 +125,9 @@ namespace :db_connector do
...
-            firstname != '' AND
-            lastname IS NOT NULL AND
-            lastname != '' AND
-            corporationname IS NOT NULL AND
-            corporationname != '' AND
-            businesshousenumber IS NOT NULL AND
-            businessstreetname IS NOT NULL AND
-            businesscity IS NOT NULL AND
-            businessstate IS NOT NULL AND
-            businesszip IS NOT NULL;"
```
rdlester, line 141: Just FYI: I'm guessing the reason this was here was as a first attempt at solving the "duplicate owners" problem. I'll post more in the Slack.

angular-rails-app/lib/tasks/db_connector.rake
```diff
@@ -156,16 +145,20 @@ namespace :db_connector do
...
-        c.statusdate, c.statusid, prob.type, prob.majorcategoryid, prob.minorcategoryid,
-        prob.codeid
+        c.statusdate, c.statusid, prob.typeid, prob.majorcategory, prob.majorcategoryid, prob.minorcategory,
+        prob.minorcategoryid, prob.code, prob.codeid, prob.statusdescription
         FROM properties AS p INNER JOIN #{ENV['MYSQL_DATABASE_DATA']}.hpd_complaints AS c
             ON p.bbl = c.bbl
         INNER JOIN #{ENV['MYSQL_DATABASE_DATA']}.hpd_complaints_problems AS prob
-            ON c.complaintid = prob.complaintid;"
+            ON c.complaintid = prob.complaintid
+        WHERE p.hpd_registration_id IN(300980, 349726, 912547, 911741);"
```
rdlester: Checking this in to master makes me a little uncomfortable. Fine in the short term, but it'd be nice to separate this "WHERE" line out into its own commit and into a "Demo" branch.

data-imports/clean_utils.py
```diff
@@ -81,7 +82,7 @@ def add_boroid(table, column):
         """.format(table=lookup_table)
     if not column_exists(table, 'bbl'):
         sql = sql + """
-        ALTER TABLE {table} ADD COLUMN {columnid} bigint(13) NOT NULL DEFAULT 0;
+        ALTER TABLE {table} ADD COLUMN {columnid} bigint(13) NULL DEFAULT NULL;
```
rdlester: Why is this changed? I just changed this to "NOT NULL" to avoid breakage of some of the data-imports.

rdlester (changes requested):
Only important thing is the change to the boroid row.

data-imports/clean_utils.py
rdlester: Hmm, there is actually an error here, but not in this line: in line 83, it should be "column_exists(table, columnid)".

nik-efimov-yyz: Hmm I never changed this not sure how it got here.

data-imports/clean_utils.py
```diff
@@ -130,7 +131,7 @@ def column_exists(table, column, schema=None):
             table_schema = '{schema}'
             AND table_name = '{table}'
             AND column_name = '{column}'
-        """.format(schema=schema, table=table, column=column)
+        """.format(schema=os.environ['MYSQL_DATABASE_DATA'], table=table, column=column)
```
rdlester: Also, this line and 116 is wrong - looks like a conflict with Chris' recent PR.

data-imports/clean_utils.py
```diff
@@ -1,5 +1,6 @@
 import os
 from utils import connect
+import os
```
rdlester: We already import os? Seems like all the changes in this file can be reverted, honestly. (Sorry for all the comments.)

nik-efimov-yyz:
@rdlester updated. After this is merged into master I'll create a demo branch with the limited import PR.

PR #173, opened 2017-05-03
Remove angular-rails-app copy from web Dockerfile

rdlester:
./angular-rails-app is already added as a volume in the docker-compose file. Copying the files over also could prevent local changes from showing up in served pages, making testing changes difficult.


chrishenry:
Were you seeing problems testing the Rails app?

We shouldn't need to do that. A volume will overlay the filesystem, resulting in any mounted volume taking precedence.

Also, should we ever decide to push the container as an artifact, we'll need that line to ensure our code is added.

Issue #112, opened 2017-03-05
Migrate HPD violations to import script


chrishenry:
Code for import is currently in https://github.com/chrishenry/deltanyc-heatseek/blob/master/JPF/mysql.skeleton.ipynb.

rdlester:
I'm on it.

rdlester:
I can't get JP's import to work in a script due to pandas errors that don't make any sense, but regardless, the old import was open violations only. We want all violations. We'll need to scrape all the zip files off of http://www1.nyc.gov/site/hpd/about/violation-open-data.page.

zip file format is defined in https://data.cityofnewyork.us/api/views/wvxf-dwi5/files/d1c8828f-bb06-45cf-a70e-01fc815b001d?download=true&filename=HPD%20Violation%20Open%20Data.pdf

aepyornis:
still need this to be done?  i have a working script for postgres (https://github.com/aepyornis/hpd-violations), probably not to hard to convert to mysql, which i could help with.

rdlester:
I've got some work in progress, I'm pretty much most of the way there, just need some time to finish, probably next week.

The extra feature I've added to the import is running the entire thing through Luigi, with individual tasks for the raw zip download / unzip and the imports. In theory, this should also provide incremental import for free.

rdlester:
But, if it's relatively straightforward to plug in your version, go ahead, then take a look at building the Rails db_connector for the violations. I'll likely swap my import script in, b/c of incremental, but the db_connector should be the same (until we switch that to run incrementally as well).

aepyornis:
@rdlester sounds good

rdlester:
See #175 for work in progress.

Issue #71, opened 2017-01-26
Incremental data import


rdlester:
Figure out scheme for incrementally updating tables with "new" data.

PR #176, opened 2017-05-30
Import HPD Violations

rdlester:
`import_hpd_violations` imports the raw violations data that is available (and in the standard csv format) from HPD, and automatically handles incremental update; `import_hpd_open_violations` imports the current open violations dump. The former is added to the luigi_master import.


rdlester:
Also ping @aepyornis

rdlester:
Fixes #112

rdlester:
First steps towards #71

data-imports/clean_utils.py
```diff
@@ -62,6 +64,10 @@ def clean_boro(table, column, replacements):
 def add_boroid(table, column):
     """ Returns a SQL statement that adds a column for boro IDs (used in BBL) and
         populates via conversion from a cleaned boro column.
+
+    Args:
+        table
+        column - Name of the column containing cleaned two-letter boro code.
```
chrishenry, line 70: 👍

data-imports/import_hpd_open_violations.py
```diff
@@ -0,0 +1,130 @@
...
+def import_csv(args):
+    csv_dir = os.path.join(BASE_DIR, table_name)
+    mkdir_p(csv_dir)
+
+    zip_file = os.path.join(csv_dir, "hpd_violations.zip")
+    csv_file = os.path.join(csv_dir, "Violation20161031.txt")
+
+    if not os.path.isfile(csv_file) or args.BUST_DISK_CACHE:
+        log.info("DL-ing HPD Buildings")
+        download_file("http://www1.nyc.gov/assets/hpd/downloads/misc/AllOpenViolations20161101.zip", zip_file)
```
chrishenry, line 98: That's the weirdest thing; why would they name a zip file with one date, but have it contain a different one.

data-imports/import_hpd_violations.py
```diff
@@ -0,0 +1,198 @@
+import calendar
+import clean_utils
+import datetime
+import luigi
+import luigi_utils
+import os
+import tempfile
+import utils
+
+class DownloadHPDViolations():
```
chrishenry: Small nitpick, but we [should be using new-style classes](https://stackoverflow.com/questions/54867/what-is-the-difference-between-old-style-and-new-style-classes-in-python) the inherit from `object`

data-imports/import_hpd_violations.py
```diff
@@ -0,0 +1,198 @@
...
+            'NOVID': 'object',
+            'NOVDescription': 'object',
+            'NOVIssuedDate': 'object',
+            'CurrentStatus': 'object',
+            'CurrentStatusDate': 'object'
+            # 'BoroID': 'object',
+            # 'Apartment': 'object',
+            # 'Story': 'object',
+            # 'ApprovedDate': 'object',
+            # 'CurrentStatusID': 'int64',
```
chrishenry, line 101: Why are these commented? Some look important, like `BoroID`

chrishenry (approved):
Looks good. I like the new class structure for hpd_violations.

data-imports/import_hpd_open_violations.py
rdlester: My guess: The zip is released on the first of the month, but the data only runs til the last day of the previous month.

data-imports/import_hpd_violations.py
rdlester: Done.

data-imports/import_hpd_violations.py
rdlester: See the comment on line 70 - they weren't included in the first v2 (txt / csv) updates from 2013, but are in the most recent updates. I didn't figure out when they started getting included. I guess I could try something like creating a method that returns different dtype_dicts for a provided date, but then we'd need to also separate database schema specification from the csv data description. Certainly doable, and maybe worth doing, given that I've been running into memory limits when trying to do stuff with this violations data.

rdlester:
I added some more comments + some fixes to the cleaning function (my computer was running out of memory when trying to run the cleaning, but I also only have 5gb left on my computer).

data-imports/clean_utils.py
```diff
@@ -75,8 +75,8 @@ def add_boroid(table, column):
     if not table_exists(lookup_table):
         sql = sql + """
         CREATE TABLE {table} (
-            boro_code varchar(2),
-            boro_id int,
+            boro_code char(2),
+            boro_id tinyint,
```
chrishenry, line 79: Nice.

data-imports/import_hpd_violations.py
chrishenry: Ah, got it.

Captured https://github.com/chrishenry/deltanyc-heatseek/issues/178

PR #179, opened 2017-06-04
Fix zip columns

rdlester:
Storing Zip codes as ints instead of strings will drop leading 0s, and will also mess up zips reported in #####-#### format.
Also, storing as strings uses less memory than bigint for plain 5-digit zips.


rdlester:
I'm just gonna go ahead and merge this, given it's a simple change. Feel free to revert if you disagree, though.

chrishenry:
👍

